PR #4, opened 2026-05-07
docs(flutter_tvos): update README for lazy init and bump version to 1.0.2

MAUstaoglu:
## Description
* Updates `flutter_tvos` README to instruct users to lazy-initialize `TvRemoteController` instead of initializing it right away in `main`.
* Removes outdated references to `runTvApp`.
* Bumps `flutter_tvos` package version to `1.0.2` and updates the CHANGELOG.
* Removes the legacy `CHANGES.md` file.


DenisovAV:
Hey @MAUstaoglu, thanks for the PR! I looked into this carefully and want to explain why I think we should keep `runTvApp` as-is.

## The binding conflict doesn't actually exist

The concern about `WidgetsFlutterBinding.ensureInitialized()` conflicting with custom bindings is understandable, but Flutter's guard makes it a non-issue in practice:

```objc
// flutter/lib/src/widgets/binding.dart
static WidgetsBinding ensureInitialized() {
  if (WidgetsBinding._instance == null) {
    WidgetsFlutterBinding();  // only runs if nothing is installed yet
  }
  return WidgetsBinding.instance;
}
```

If a user already called `MyCustomBinding.ensureInitialized()` before `runTvApp`, the call inside `runTvApp` is a no-op — it returns the existing instance. No conflict. I traced this against Flutter 3.32.8 source to confirm.

Additionally, `TvRemoteController.init()` only needs `ServicesBinding` (any custom binding includes it) — it has no hard dependency on `WidgetsFlutterBinding` specifically. The channel's `setMessageHandler` bottoms out in `ui.channelBuffers.setListener()`, which is a dart:ui engine primitive independent of which binding subclass is installed.

## Lazy init is a regression

The lazy initialization approach (registering the channel handler on first use) has a real failure mode: if the Siri Remote is pressed before any listener is added, the event arrives in the channel buffer and is **silently lost**. `runTvApp` prevents this by guaranteeing the handler is registered before `runApp` builds the first frame.

## `runTvApp` is already the simplest possible API

Asking users to call `TvRemoteController.instance.init()` manually trades one word (`runApp` → `runTvApp`) for two mandatory, order-sensitive calls with no compiler enforcement. This is a net regression in DX.

The one-liner pattern is well-established in the ecosystem — Sentry's `SentryFlutter.init`, Riverpod's `ProviderScope`, etc. all follow the same "wrap your entry point" approach.

## Verdict

I'd like to close this PR and keep `runTvApp` as the primary API. Happy to add a note to the docs clarifying why the binding conflict is not a concern, if that helps.

DenisovAV (approved):

PR #11, opened 2026-05-24
flutter-tvos v1.1.0 — the porter release

### LINKED ISSUE #3: Target native_assets required define SdkRoot but it was not provided
Hello,

I'm trying to migrate an application to flutter-tvos.

I started from a fresh project and added dependencies one by one.

As expected, packages without tvOS support do not build. I then followed the documented steps to add a tvOS implementation to one of the packages (path_provider), but after doing so, the app fails to build with the following error:

`Target native_assets required define SdkRoot but it was not provided `

At this point I'm not sure whether:
- I made a mistake while creating the tvOS package
- there is something wrong in flutter-tvos

Is this something you already encountered, or do you know how I could fix it?

MAUstaoglu:
Minor release — the "porter release". `v3.41.9-tvos.1.0.1` had no
`lib/plugin_porting/` or `flutter-tvos plugin port` command at all;
this PR ships the entire 7-phase porter for scaffolding federated
`*_tvos` plugins from existing iOS / macOS plugins, plus first-class
`flutter-tvos create --platforms=tvos`, a build-time nudge for
plugins missing tvOS support, and a handful of build-pipeline fixes.

No Flutter SDK or engine-artefact change — pinned versions match
`v1.0.1` (Flutter `3.41.9` / `00b0c91f06`, tvOS engine artefacts
`v1.0.0-flutter3.41.9`). Release tag will be `v3.41.9-tvos.1.1.0`
per the existing `v<flutter-version>-tvos.<x.y.z>` scheme.

### Added

- **`flutter-tvos plugin port` — new command.** Scaffolds a
  federated `<plugin>_tvos` sibling from any existing iOS or macOS
  Flutter plugin. Source loaders: positional local path,
  `--from-pub <name>`, or `--from-git <url> --ref <ref>`. The 11
  packages at [`fluttertv/plugins`](https://github.com/fluttertv/plugins)
  / [pub.dev under fluttertv.dev](https://pub.dev/publishers/fluttertv.dev/packages)
  were all produced this way. The transformer:
  - widens Swift `#if os(iOS)` / `#elseif os(macOS)` and ObjC
    `#if TARGET_OS_IOS` so tvOS follows the iOS branch;
  - widens `@available iOS X, *` / `#available` /
    `API_AVAILABLE(ios())` to also list `tvOS X`;
  - matches a compatibility DB of tvOS-incompatible APIs (WebKit,
    SafariServices, LocalAuthentication, CoreLocation,
    CaptiveNetwork, NEHotspot, StoreKit code-redemption,
    `UIPasteboard`, AVAudioSession Bluetooth / speaker options,
    CoreTelephony, GoogleSignIn SDK, …) and either strips the
    import + stubs the enclosing method-channel handler, or wraps
    type-level uses behind `#if !os(tvOS)` (graceful partial port);
  - collapses modern multi-target SwiftPM packages into one
    CocoaPods module;
  - emits a buildable Swift skeleton for FFI / native-assets
    plugins the tvOS toolchain can't build for as-is;
  - prunes cross-platform Dart from the generated `lib/` (drops
    `_plus`-style packages' Linux / Windows / Web / macOS /
    Android implementations + scrubs their imports).

  Every transformation is recorded in a `PORTING_REPORT.md`
  written alongside the package.

- **`flutter-tvos create --platforms=tvos` / `--tvos-only`** —
  genuine tvOS-only project scaffold (no more
  create-iOS-then-strip).

- **Build-time warning for plugins missing tvOS support.** Each
  plugin in the app's dep graph that has a FlutterTV-published
  `<name>_tvos` sibling the user hasn't added is surfaced as a
  one-line warning during `flutter-tvos build/run`. Keyed on the
  user-facing aggregator name so `_darwin` / `_android` /
  `_linux` siblings don't produce duplicate or noisy warnings.

- **New `doc/port-plugin.md`** — proper user guide for `plugin
  port` (quick start, full flag reference, what the transformer
  does and doesn't do, after-porting workflow, troubleshooting).

### Fixed

- `Generated.xcconfig` now reads `FLUTTER_BUILD_NAME` /
  `FLUTTER_BUILD_NUMBER` from the app's `pubspec.yaml` `version:`
  (`1.2.3+4` → `1.2.3` / `4`). Previously hardcoded to `1.0.0` /
  `1` unless overridden via `--build-name` / `--build-number`,
  which broke `package_info_plus` and any code reading
  `CFBundleShortVersionString` / `CFBundleVersion`.
- `FLTAssetsPath` is now baked into the generated `Info.plist` so
  plugin asset lookup resolves on a real Apple TV (not just the
  simulator).
- Device builds now pass `-allowProvisioningUpdates` to xcodebuild
  so the first build on a fresh team / device can refresh
  provisioning profiles itself.
- tvOS build skips the Dart native-assets target so apps that
  transitively pull in an FFI / native-assets plugin (e.g.
  `path_provider_foundation`, `package_info_plus`) no longer fail
  with `Target native_assets required define SdkRoot` on first
  build (issue #3).

### Documentation

- README's plugin-porting section is now a 4-line pointer to
  `doc/port-plugin.md`; the platform-key paragraph points at the
  published [`fluttertv/plugins`](https://github.com/fluttertv/plugins)
  repo and the [`fluttertv.dev`](https://pub.dev/publishers/fluttertv.dev/packages)
  verified publisher in place of "FlutterTV-curated index being
  assembled and will be published soon".
- `doc/` index gains "Porting an existing plugin" under "Plugin
  development".

## Verification

- `dart analyze lib/` — 0 errors, 0 warnings
- `dart test test/general/` — 191 / 191 passing
- Built and verified end-to-end: scaffolded the 11 packages at
  [`fluttertv/plugins`](https://github.com/fluttertv/plugins) via
  `--from-pub`, published them to pub.dev under
  [`fluttertv.dev`](https://pub.dev/publishers/fluttertv.dev/packages),
  and exercised every one in a sample app on the Apple TV
  simulator.

See [`CHANGELOG.md`](CHANGELOG.md) for the full per-line entry.

---

Closes #3.


lib/plugin_porting/scaffolder.dart
```diff
@@ -0,0 +1,702 @@
...
+    // One regex covers `import 'X';`, `export 'X';`, and the conditional
+    // form with up to one `if (...) 'Y'` fork.
+    final RegExp directive = RegExp(
+      r"""^(\s*)(import|export)\s+(['"])([^'"]+)\3"""
+      r"""(\s+if\s*\(\s*[^)]*\)\s+(['"])([^'"]+)\6)?"""
+      r"""([^;\n]*);?\s*$""",
+      multiLine: true,
+    );
+
+    return source.replaceAllMapped(directive, (Match m) {
```
DenisovAV, line 490: `part` directives are not matched — only `import` and `export` are handled here.

A file containing `part 'src/foo_io.dart';` pointing to a dropped path will survive unchanged in the generated output, producing a broken reference the Dart compiler will reject. This is rare in modern plugins but real in some older or Pigeon-heavy codebases.

Fix: extend the regex to also cover `part`:
```dart
final RegExp directive = RegExp(
  r"""^(\s*)(import|export|part)\s+(['"])([^'"]+)\3""",
  multiLine: true,
);
```
`part of 'parent.dart';` directives reference the parent file and should not be scrubbed — only the bare `part 'child.dart';` form.

lib/plugin_porting/swift_porter.dart
```diff
@@ -0,0 +1,601 @@
...
+  /// `@available(...)` attribute or `#available(...)` runtime check.
+  /// Group 1 = the keyword (`@available`/`#available`), group 2 = the
+  /// platform list between the parentheses.
+  static final RegExp _availabilityClause =
+      RegExp(r'(@available|#available)\s*\(([^)]*)\)');
+
+  /// An `iOS <version>` entry inside an availability clause. Group 1 is
+  /// the version (`15`, `15.0`, `17.4`, …). The leading boundary avoids
+  /// matching the `iOS` inside identifiers (there are none in practice,
+  /// but be safe).
```
DenisovAV, line 35: `[^)]*` does not cross newlines, so multi-line `@available` attributes are silently skipped without widening.

This pattern appears in Pigeon-generated Swift for complex platform lists:
```swift
@available(iOS 15.0,
           macOS 12.0, *)
func foo() {}
```

The line `@available(iOS 15.0,` fails to match the full clause, so no `tvOS 15.0` entry is inserted. The port compiles fine on iOS but produces an "unavailable" error on tvOS for the symbol gated by that attribute.

Fix: use `RegExp(r'(@available|#available)\s*\(([^)]*\)', dotAll: true)` and process the full source string rather than line-by-line for this pass, or use a two-pass approach that first joins continuation lines before applying the clause regex.

lib/plugin_porting/compatibility_database.dart
```diff
@@ -0,0 +1,313 @@
...
+  ///
+  /// The porter does an exact-line match against trimmed source lines,
+  /// so each entry should be the literal `import` directive.
+  final List<String> stripImports;
+}
+
+/// The database. Append-only; existing entries should not be removed when
+/// a tvOS API gap closes — instead set the severity to [Severity.info] and
+/// note the version it became available, so users porting on older OS
+/// targets still see the warning.
```
DenisovAV, line 88: The `ObjcPorter` derives framework names from `stripImports` by stripping the `import ` prefix (`imp.substring(7).trim()`), so this field implicitly assumes Swift import syntax.

An entry like `stripImports: ['#import <LocalAuthentication/LocalAuthentication.h>']` would silently produce the wrong key in `_bannedFrameworks` and the ObjC import would never be commented out — with no test or runtime error to surface the problem.

All current entries already use Swift form (`import WebKit`, etc.), so nothing is broken today. But the assumption is invisible at the type boundary.

Suggest either:
1. Rename the field to `stripSwiftImports` to make the Swift-only contract explicit, or
2. Add a constructor assert: `assert(stripImports.every((s) => s.startsWith('import ')), 'stripImports entries must use Swift import syntax')`

DenisovAV (approved):
All three issues fixed — `part` directive scrubbing, multi-line `@available` via `dotAll: true`, and `stripImports` renamed to `stripSwiftImports`. LGTM.

PR #17, opened 2026-05-28
fix(build): honour --obfuscate and --split-debug-info in tvOS AOT builds

### LINKED ISSUE #10: obfuscate flag fails silently
Running `flutter-tvos build tvos --obfuscate --split-debug-info=<path>` completes without errors but the flags have no effect. The output App.framework/App binary retains fully readable Dart symbols, as if the flags were never passed.

The same project and Dart codebase produces a correctly obfuscated binary on iOS using the standard Flutter toolchain, so this appears to be specific to flutter-tvos. 

Expected: symbols obfuscated, debug info written to the specified path.                              
Actual: binary unchanged, debug info path empty.

MAUstaoglu:
## Summary
- The tvOS AOT step runs `gen_snapshot` directly (not upstream `AOTSnapshotter`) and previously dropped `--obfuscate` / `--split-debug-info` entirely — release/profile builds completed but the `App` binary kept readable Dart symbols and the split-debug-info directory stayed empty.
- The `gen_snapshot` invocation now forwards `--obfuscate`, the `--dwarf-stack-traces` / `--resolve-dwarf-paths` / `--save-debugging-info=<dir>/app.ios-arm64.symbols` trio, and any `--extra-gen-snapshot-options`, matching `AOTSnapshotter.build`.
- Arg construction extracted into a testable `tvosGenSnapshotArgs` helper; bumped to 1.1.1.

## Test plan
- [x] `dart analyze --fatal-infos` clean on changed files
- [x] New unit tests (`test/general/tvos_aot_snapshot_test.dart`) cover the flag matrix — 8/8 passing
- [x] Manual A/B release build of a demo app: app Dart symbols drop from present to 0 hits with the flags, and `app.ios-arm64.symbols` is written

Fixes #10

### DIFF
--- a/lib/build_targets/application.dart
+++ b/lib/build_targets/application.dart
@@ -12,6 +12,7 @@ import 'package:flutter_tools/src/build_system/targets/common.dart';
 import 'package:flutter_tools/src/build_system/targets/localizations.dart';
 import 'package:flutter_tools/src/globals.dart' as globals;
 import 'package:flutter_tools/src/project.dart';
+import 'package:meta/meta.dart';
 
 import '../tvos_artifacts.dart';
 import '../tvos_build_info.dart';
@@ -706,14 +707,22 @@ class NativeTvosBundle extends Target {
 
     final String assemblyPath = globals.fs.path.join(aotDir.path, 'snapshot_assembly.S');
 
+    // The --split-debug-info directory must exist before gen_snapshot writes to it.
+    final String? splitDebugInfo = environment.defines[kSplitDebugInfo];
+    if (splitDebugInfo != null && splitDebugInfo.isNotEmpty) {
+      globals.fs.directory(splitDebugInfo).createSync(recursive: true);
+    }
+
     // Run gen_snapshot to produce assembly
-    final ProcessResult genSnapshotResult = await globals.processManager.run(<String>[
-      genSnapshotPath,
-      '--deterministic',
-      '--snapshot_kind=app-aot-assembly',
-      '--assembly=$assemblyPath',
-      kernelSnapshot.path,
-    ]);
+    final ProcessResult genSnapshotResult = await globals.processManager.run(
+      tvosGenSnapshotArgs(
+        fileSystem: globals.fs,
+        genSnapshotPath: genSnapshotPath,
+        assemblyPath: assemblyPath,
+        kernelSnapshotPath: kernelSnapshot.path,
+        defines: environment.defines,
+      ),
+    );
 
     if (genSnapshotResult.exitCode != 0) {
       globals.logger.printError('gen_snapshot failed:');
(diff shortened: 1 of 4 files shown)


test/general/tvos_aot_snapshot_test.dart
```diff
@@ -0,0 +1,94 @@
...
+
+  testWithoutContext('omits --obfuscate and debug flags when defines are absent', () {
+    final List<String> args = argsFor(<String, String>{});
+    expect(args, isNot(contains('--obfuscate')));
+    expect(args, isNot(contains('--dwarf-stack-traces')));
+    expect(args.any((String a) => a.startsWith('--save-debugging-info=')), isFalse);
+  });
+
+  testWithoutContext('passes --obfuscate when kDartObfuscation is true', () {
+    final List<String> args = argsFor(<String, String>{kDartObfuscation: 'true'});
```
DenisovAV: This test exercises a combination that is unreachable in production: upstream `FlutterCommand.getBuildInfo()` calls `throwToolExit` when `--obfuscate` is passed without `--split-debug-info`. The test therefore gives false coverage confidence.

Two options:
1. Delete this test case (the unreachable path is guarded by the CLI layer, not by `tvosGenSnapshotArgs` itself).
2. Add a note that this tests the helper in isolation and the CLI-level guard makes it unreachable, then invert the assertion to match what the helper *should* do — i.e., not emit `--obfuscate` when `shouldSplitDebugInfo` is false (mirroring `AOTSnapshotter` which conditions `--obfuscate` on `if (shouldSplitDebugInfo)`).

Either way, the `assert(args, contains("--obfuscate"))` here is testing behaviour that would be rejected before reaching `tvosGenSnapshotArgs`.

lib/build_targets/application.dart
```diff
@@ -810,6 +819,47 @@ class NativeTvosBundle extends Target {
...
+    final dartObfuscation = defines[kDartObfuscation] == 'true';
+    final String? splitDebugInfo = defines[kSplitDebugInfo];
+    final bool shouldSplitDebugInfo = splitDebugInfo != null && splitDebugInfo.isNotEmpty;
+    final List<String> extraGenSnapshotOptions = decodeCommaSeparated(
+      defines,
+      kExtraGenSnapshotOptions,
+    );
+    final String? saveDebuggingInfoArg = shouldSplitDebugInfo
+        ? '--save-debugging-info=${fileSystem.path.join(splitDebugInfo, 'app.ios-arm64.symbols')}'
+        : null;
```
DenisovAV, line 846: The symbols filename is hardcoded as `app.ios-arm64.symbols` — the `ios-arm64` suffix comes from iOS naming conventions in `AOTSnapshotter`, but this is a tvOS build. A user trying to symbolicate a tvOS crash will look for `tvos-arm64.symbols` (or whatever `otool -l` reports as the platform) and won't find it.

Upstream computes the name dynamically via `getNameForTargetPlatform(platform, darwinArch: darwinArch)`. Worth using `tvos-arm64` here to avoid confusion when following Xcode crash report symbolication guides.

DenisovAV (approved):
Both issues addressed: symbols file renamed to app.tvos-arm64.symbols, and the obfuscate-only test documented with the CLI-layer guard explanation. LGTM.

PR #19, opened 2026-06-05
Upgrade to Flutter 3.44.1 + native tvOS shaders + wireless on-device debug

MAUstaoglu:
## Summary

Bumps the pinned Flutter SDK to **3.44.1** (`924134a44c189315be2148659913dda1671cbe99`, Dart 3.12.1) and the engine artifacts to `v1.0.0-flutter3.44.1`, plus two robustness wins that come with the new engine:

1. **Impeller Metal shaders are now tvOS-native** in the engine artifact — retires the per-app `MetalLibInterposer` hack.
2. **On-device debug + hot reload + DevTools now work over a wireless Apple TV**, by mirroring stock Flutter iOS's lldb → Xcode-debugger launch flow.

Engine-side changes ship in [`fluttertv/engine#1`](https://github.com/fluttertv/engine/pull/1); artifacts at [`fluttertv/engine-artifacts@v1.0.0-flutter3.44.1`](https://github.com/fluttertv/engine-artifacts/releases/tag/v1.0.0-flutter3.44.1).

### Native tvOS Impeller shaders (interposer removed)
- Impeller's `compute` / `entity` / `framebuffer_blend` / `modern` metallibs are compiled with the `appletvos` SDK inside the engine and embedded in `Flutter.framework`, loading directly on the Apple TV GPU like iOS.
- Removes the per-app `tvos_metallibs/` directory and `MetalLibInterposer` (which swizzled `newLibraryWithData:` to swap shaders matched by exact byte size — coupled to the Xcode/Metal-compiler version, so it broke on engine rebuilds) from the app template and the `flutter_tvos` example, including the bridging-header entry.

### Wireless on-device debugging (like stock iOS)
- `TvosDevice` now mirrors `IOSDevice._startAppOnCoreDevice`: attach lldb first, and when the wireless CoreDevice tunnel stalls ("LLDB is taking longer than expected") or drops, **fall back to the Xcode debugger** (`XcodeDebug.debugApp` via AppleScript against the generated `tvos/Runner.xcworkspace` + `Runner` scheme). The VM Service is then resolved over mDNS at the device's LAN IP, so hot reload (`r`), hot restart (`R`), and DevTools (`d`) come for free from the base `HotRunner`.
- First run prompts to allow controlling Xcode (Settings ▸ Privacy & Security ▸ Automation) — same as stock iOS.

### Version / docs
- `bin/internal/flutter.version` → `924134a…`; `bin/internal/engine.version` → `v1.0.0-flutter3.44.1`.
- Bump to **1.2.0**; refresh `CHANGELOG.md`, `README.md`, `doc/architecture.md`, `doc/get-started.md`.

### Tests
- Added `TvosDevice.parseDeviceUdid` coverage (extract UDID from `devicectl` JSON, null on missing / malformed).

## Compatibility

**Existing apps keep working without changes.** Apps built from an older template still ship a `MetalLibInterposer`; against the 3.44.1 engine its byte-size-keyed swizzle finds no match and goes inert — the engine's own tvOS-native shaders are used. To drop the now-dead files, recreate the `tvos/` runner or delete `Runner/MetalLibInterposer.*`, `Runner/tvos_metallibs/`, and the bridging-header `#import`.

## Verification

- `dart analyze` clean on changed `lib/` files; `tvos_physical_device_test.dart` 21/21 passing.
- Validated on the tvOS Simulator and a physical Apple TV across debug / profile / release, including hot reload and DevTools over wireless.


lib/tvos_device.dart
```diff
@@ -742,6 +795,153 @@ class TvosDevice extends Device {
...
+        'Apple TV via USB for lldb debugging.',
+      );
+      return false;
+    }
+
+    final xcodeDebug = XcodeDebug(
+      logger: logger,
+      processManager: globals.processManager,
+      xcode: globals.xcode!,
+      fileSystem: globals.fs,
```
DenisovAV, line 927: This force-unwraps `globals.xcode!`, but this fallback fires precisely when something already went wrong (lldb did not attach). On a machine where Xcode is not selected (`xcode-select` points at Command Line Tools only, or a fresh CI box), `globals.xcode` is null and `!` throws a bare `Null check operator used on a null value` — and by this point `_teardownDeviceLaunch()` has already killed the working devicectl+lldb launch, so there is no recovery.

Note upstream `IOSDevice` passes the **nullable** `globals.xcode` here (`devices.dart` → `xcode: globals.xcode`), so the `!` is a regression vs the iOS pattern, not parity.

**Fix:** guard `globals.xcode == null` → `printError` an actionable message ("Xcode is required for the wireless debug fallback; open Xcode once / run `xcode-select`") and `return false`.

lib/tvos_device.dart
```diff
@@ -1081,6 +1281,8 @@ class TvosDevice extends Device {
     _lldb = null;
     unawaited(_lldbLogForwarder?.exit());
     _lldbLogForwarder = null;
+    unawaited(_xcodeDebug?.exit());
+    _xcodeDebug = null;
```
DenisovAV, line 1407: Good — but this teardown only runs in `stopApp()`. `dispose()` (which runs on abnormal teardown, `flutter-tvos attach` exit, and hot-restart) only disposes the log reader and does **not** call `_xcodeDebug?.exit()`. If the Xcode fallback was active when `dispose()` runs, the `osascript` automation process leaks and the debug session stays attached on the Apple TV.

The base code already cleans `_lldb` / `_lldbLogForwarder` in `stopApp`; the symmetry for `dispose` is missing.

**Fix:** mirror this in `dispose()` — `unawaited(_xcodeDebug?.exit()); _xcodeDebug = null;`.

lib/tvos_device.dart
```diff
@@ -637,36 +639,87 @@ class TvosDevice extends Device {
...
+        );
+        if (!xcodeStarted) {
+          logger.printError('Failed to start a debug session on the device.');
+          return LaunchResult.failed();
+        }
+        // Xcode launched the app; its console output isn't routed through our
+        // log reader, so resolve the VM Service purely via mDNS (Bonjour) using
+        // the device's LAN IP — exactly how stock iOS resolves it for wireless.
+        final Uri? xcodeUri = await MDnsVmServiceDiscovery.instance!.getVMServiceUriForAttach(
+          bundleId,
```
DenisovAV: After `_launchViaXcodeDebugger` succeeds, `getVMServiceUriForAttach` can `throwToolExit` — multiple Dart VM services on the LAN (`mdns_discovery.dart` `queryForAttach`) or denied macOS Local Network permission (`throwOnMissingLocalNetworkPermissionsError` defaults to `true`). At this point `_xcodeDebug` is set, there is no try/finally around this block, and there is no shutdown hook (upstream `_startAppOnCoreDevice` adds one). So the throw exits `_startAppOnDevice` leaving an **orphaned Xcode debug session attached on the device**.

**Fix:** wrap the Xcode-launch + mDNS-resolve block in try/finally that runs `_xcodeDebug.exit()`, or pass `throwOnMissingLocalNetworkPermissionsError: false` and handle the null return (see the sibling comment on the `succeeded()` line).

lib/tvos_device.dart
```diff
@@ -637,36 +639,87 @@ class TvosDevice extends Device {
...
+          this,
+          useDeviceIPAsHost: true,
+          timeout: const Duration(seconds: 60),
+        );
+        if (xcodeUri != null) {
+          logger.printTrace('VM service (via Xcode + mDNS) available at: $xcodeUri');
+          return LaunchResult.succeeded(vmServiceUri: xcodeUri);
+        }
+        return LaunchResult.succeeded();
       }
```
DenisovAV, line 788: When mDNS yields no `xcodeUri` (VM service never advertised within 60s), this returns `LaunchResult.succeeded()` with **no** `vmServiceUri`. Traced through `resident_runner.dart`: `started == true && hasVmService == false` → `vmServiceUris = Stream.empty()` → the runner reports success and waits forever. Net effect: the app launches, the CLI says success, but **hot reload, hot restart, and DevTools are silently non-functional with zero diagnostic output** — the most user-hostile outcome of the new flow.

**Fix:** before returning the bare `succeeded()`, emit `printWarning`: "App launched via Xcode but the Dart VM Service was not found over mDNS within 60s — hot reload and DevTools will be unavailable. Check the Mac Local Network permission and that the Apple TV is on the same LAN." (The non-fallback tail a few lines below copies the same anti-pattern — worth fixing both.)

lib/tvos_device.dart
```diff
@@ -742,6 +795,153 @@ class TvosDevice extends Device {
...
+
+    // Ensure the Runner scheme has a debugger launch action (Xcode needs it to
+    // attach the debugger when it runs the scheme).
+    final File schemeFile = xcodeproj
+        .childDirectory('xcshareddata')
+        .childDirectory('xcschemes')
+        .childFile('Runner.xcscheme');
+    if (schemeFile.existsSync()) {
+      xcodeDebug.ensureXcodeDebuggerLaunchAction(schemeFile);
+    }
```
DenisovAV, line 952: `ensureXcodeDebuggerLaunchAction` calls `throwToolExit(...)` upstream when the Runner scheme LaunchAction does not select the LLDB debugger (the upstream comment notes users have reported schemes not being so). Called unguarded here, a project whose scheme is not debugger-configured aborts the entire `run` instead of returning `false` and printing the intended "Failed to start a debug session" message.

It is read-only validation (does not mutate the scheme, so no restore-cleanup needed) — the only gap is the unguarded throw.

**Fix:** wrap in try/catch → `return false`.

lib/tvos_device.dart
```diff
@@ -742,6 +795,153 @@ class TvosDevice extends Device {
...
+        'device',
+        'info',
+        'details',
+        '--device',
+        deviceId,
+        '--json-output',
+        out.path,
+      ]);
+      if (r.exitCode != 0 || !out.existsSync()) {
+        return null;
```
DenisovAV, line 1029: This branch returns `null` with **no log of the devicectl exit code or stderr**, and the caller does `await _resolveDeviceUdid(id) ?? id` — so on failure it silently hands the CoreDevice **GUID** (which Xcode does not recognize) to `debugApp`. The user then gets a confusing "device not found in Xcode" much later, with no breadcrumb that UDID resolution actually failed.

**Fix:** `printTrace` the `r.exitCode` and `r.stderr` in this branch, and `printTrace` at the call site when falling back to the raw `id`. (Same idea for `parseDeviceUdid` returning null on `exitCode == 0` — a devicectl JSON-shape drift is currently invisible; a trace there distinguishes "no UDID" from "schema changed".)

DenisovAV (changes requested):
Reviewed with 6 parallel reviewers (CLI, build/shaders, wireless-debug lifecycle, architecture, silent-failure, type-design), cross-checked against the pinned 3.44.1 SDK. Full report: `docs/pr-reviews/pr-19-review.md`.

## The clean parts — ready as-is
- **MetalLibInterposer removal** is complete and symmetric across example + template: every `PBXFileReference` removed with its `PBXBuildFile` and build-phase membership, no orphaned IDs, all four `.metallib` binaries gone, bridging header edited-not-deleted (still valid). The "goes inert against the 3.44.1 engine" claim holds.
- **`application.dart` rewrite** — no regression to PR #17 obfuscate/split-debug-info work. The `Embed App.framework` `rsync --delete` was empirically verified not to nuke sibling `Flutter.framework`.
- **Version bump** consistent (3.44.1 / `v1.0.0-flutter3.44.1` / 1.2.0). 3.44.1 is a stable upstream patch, so the bump is the right call.
- Wireless flow **faithfully mirrors upstream `_startAppOnCoreDevice`**; AppleScript is injection-safe (separate argv), mDNS client closed in `finally`, and the 75s timeout for stall detection is actually **more robust** than upstream string-matching.

## Requesting changes — all in the wireless-debug error/cleanup edges
Six inline comments above (priority order):

1. **Silent dead hot reload** — `succeeded()` with no VM service URI reports success but leaves hot reload / DevTools non-functional with zero output. Needs a `printWarning`.
2. **Orphaned Xcode session** — no try/finally around Xcode-launch + mDNS; a `throwToolExit` from mDNS (e.g. denied Local Network permission) leaks the attached debug session.
3. **`globals.xcode!`** force-unwrap crashes if Xcode is not selected (upstream passes the nullable).
4. **`ensureXcodeDebuggerLaunchAction`** unguarded `throwToolExit` aborts the whole `run`.
5. **`_xcodeDebug` not cleaned up in `dispose()`** (only `stopApp`) — leaks on abnormal teardown.
6. **`_resolveDeviceUdid`** swallows devicectl stderr, then silently feeds Xcode the wrong identifier.

## Docs (not in the diff, please update before tagging)
- `README.md` still says **"Debug mode is simulator-only … Debug (JIT) is blocked on the device by Apple"** — directly contradicted by this PR.
- `doc/architecture.md` Build Flow still lists the removed "Copy Metal libraries" step and the Run Flow does not mention the Xcode fallback.

Plus 11 minor items in the report (omitted `expectedConfigurationBuildDir`, log-forwarder subscription leak, unguarded temp-dir delete, example `DEVELOPMENT_TEAM`, `_kTvosMinimumOSVersion` not asserted against the template, etc.) — nice-to-have, not blocking.

The logic is sound and mirrors upstream well — these are all small guards / `printWarning` / try-finally, no rework needed.

MAUstaoglu:
Thanks for the thorough review — all six points addressed in `fe2050b` (all in `lib/tvos_device.dart`):

1. **`globals.xcode!` force-unwrap** → now `final Xcode? xcode = globals.xcode;` with a null guard that `printError`s an actionable message ("Xcode is required … open Xcode / `xcode-select` …") and returns `false`, matching upstream IOSDevice's nullable handling.
2. **`dispose()` leaked `_xcodeDebug`** → added `unawaited(_xcodeDebug?.exit()); _xcodeDebug = null;` to `dispose()`, mirroring `stopApp()`.
3. **Orphaned Xcode session on mDNS throw** → the post-launch `getVMServiceUriForAttach` is now wrapped in try/catch, so a throw (multiple VM services / denied Local Network permission) no longer escapes and crashes `run` leaving the session attached.
4. **Silent no-VM-Service success** → both the Xcode-fallback path *and* the normal tail now `printWarning` (Local Network permission + same-LAN hint) before the bare `succeeded()`, so hot reload/DevTools being unavailable isn't mute.
5. **Unguarded `ensureXcodeDebuggerLaunchAction` throw** → wrapped in try/catch → `printError` + `return false` ("Failed to start a debug session") instead of aborting the run.
6. **Silent UDID-resolution failure** → `_resolveDeviceUdid` now `printTrace`s the devicectl exit code/stderr and the "JSON shape changed" case, and the call site traces the raw-id fallback.

`dart analyze` clean; full suite 214 tests pass. Ready for another look when you have a chance.

MAUstaoglu:
Added `9b49005`: on-device wireless debug is now actually reliable, paired with the engine fix.

Root cause of the earlier flakiness was an **engine** gap, not the CLI: the tvOS debug engine was missing the `NOTIFY_DEBUGGER_ABOUT_RX_PAGES` Dart-VM hook that lldb's attach breakpoints, so the attach hung and fell back to the Xcode debugger (which doesn't sustain a session on tvOS). Fixed in fluttertv/engine#1; the rebuilt `tvos_debug_arm64` artifact is now in the `v1.0.0-flutter3.44.1` release.

CLI side in this commit:
- lldb attach timeout 75s → 180s, overridable via `FLUTTER_TVOS_LLDB_ATTACH_TIMEOUT_SECONDS` (Apple TV is wireless-only, so the attach always goes over the slower CoreDevice tunnel).
- Actionable failure message (restart the Apple TV to reset the tunnel / check Local Network permission / use the simulator) instead of a bare "Failed to start a debug session".

Verified end-to-end on a physical Apple TV (tvOS 18.6): lldb attaches, hot reload ×2 + hot restart + DevTools all work, no Xcode fallback. Note for existing checkouts: `flutter-tvos precache --force` to pick up the updated engine artifact.

DenisovAV:
Verified locally on `9b49005b` against the 3.44.1 SDK — `dart analyze` clean (0 errors/warnings) and **214/214 tests pass**. Confirmed all six review points are addressed in `fe2050b`, and `9b49005b` (generous lldb timeout + honest failure UX, `_teardownDeviceLaunch()` before the Xcode fallback) is a solid follow-up with no new issues. Nice work.

One thing left before I approve — the **docs** point from the original review is still open. Two stale spots that now contradict the feature this PR ships:

1. **`README.md:146`** still says:
   > **Debug mode is simulator-only.** Physical Apple TV deployment runs in release/profile mode (AOT). Debug (JIT) is blocked on the device by Apple.

   This is now false — the PR adds on-device wireless debug. Please update it to reflect that debug + hot reload + DevTools now work over a wireless Apple TV (with the engine artifact bundled here), while noting the simulator remains the fast-iteration path.

2. **`doc/architecture.md:96`** still lists the build step:
   > **Copy Metal libraries** — copies pre-compiled tvOS Metal shaders from templates

   This step was removed in this PR (shaders are now tvOS-native in the engine). Please drop it, and ideally add the Xcode-debugger fallback to the Run Flow section.

Both are quick text edits. Once they land I will approve and we can merge for the 1.2.0 release. Thanks!

MAUstaoglu:
Both docs points landed in `b6d3dc1`:

1. **`README.md`** — replaced the stale *"Debug mode is simulator-only … Debug (JIT) is blocked on the device by Apple"* bullet with the real flow: on-device wireless debug now gives hot reload + hot restart + DevTools over the CoreDevice tunnel (the bundled tvOS debug engine ships the `NOTIFY_DEBUGGER_ABOUT_RX_PAGES` hook lldb needs), while calling out that the **simulator stays the fast-iteration path** since the wireless attach is slower than USB.
2. **`doc/architecture.md`** — dropped the removed *"Copy Metal libraries"* build step (shaders are now compiled tvOS-native inside the engine artifact), and added a **Physical Apple TV (debug)** subsection to the Run Flow describing the lldb → Xcode-debugger fallback and mDNS VM-service discovery.

Also grepped the docs tree for any other `metallib` / `Interposer` / `simulator-only` leftovers — none remain. Ready for another look, thanks!

DenisovAV (approved):
Docs updated correctly (`b6d3dc1a`, README + architecture.md only, no code): the false "debug is simulator-only" line is replaced with an accurate on-device wireless-debug description, and the removed "Copy Metal libraries" build step is dropped — plus a proper lldb → Xcode-fallback → mDNS Run Flow section.

Re-verified the whole PR: `dart analyze` clean, 214/214 tests pass, all six review points fixed in `fe2050b`, hardening in `9b49005b` sound. Approving for the 1.2.0 release. Thanks for the thorough turnaround.

PR #23, opened 2026-06-12
Release 1.3.0 — Swift Package Manager support + flutter-tvos upgrade

MAUstaoglu:
## Release 1.3.0

Promotes `dev` to `main` for the **flutter-tvos 1.3.0** release. Minor release — new features, no breaking changes, **no engine or Flutter SDK change** (pinned versions match 1.2.0: Flutter 3.44.1, engine `v1.0.0-flutter3.44.1`).

### What's in it
- **Swift Package Manager support for tvOS apps (default; CocoaPods kept).** Newly created apps link a generated `FlutterGeneratedPluginSwiftPackage` umbrella (`FlutterFramework` binary target + every plugin shipping a `tvos/Package.swift`), regenerated each build and wired into the Runner project (`objectVersion = 56`). Pod-only plugins keep resolving through CocoaPods; the Podfile skips any plugin with a `Package.swift` so the two never double-link. Verified end-to-end with five real federated plugins on the simulator (debug) and a release (AOT) build on a physical Apple TV.
- **`flutter-tvos upgrade`** — upgrades the toolchain to its newest release tag (re-pins Flutter + engine artifacts together), with `--verify-only` / `--force`.
- **`flutter_tvos` package → 1.1.0** gains SPM support (FFI shim links statically through the umbrella; CocoaPods unchanged).
- CI runs on `dev` and dev PRs.

### Release-prep in this PR
- `version` 1.2.0 → **1.3.0**, CHANGELOG finalized, README current-version updated.

### After merge
Tag `main` with **`v3.44.1-tvos.1.3.0`** → the Release workflow publishes the GitHub release. Then `flutter_tvos 1.1.0` and the five `*_tvos` 0.0.2 packages can go to pub.dev.

### Verification
- `dart analyze` clean, **242 tests pass**.
- Backward compat checked: old-style projects + new CLI still resolve plugins via CocoaPods (no regression).


DenisovAV:
## Review — multi-agent pass + adversarial verification

Ran a 6-reviewer pass over the diff, then a second adversarial pass that re-verified every Critical/Important finding against the actual code at `dev` (4a84145). Findings that didn't survive verification were downgraded — noted below. Line refs are at the PR head.

**Recommendation: request changes** — 1 critical, 6 important. The critical one is a five-line fix.

---

### Critical

**C1. `_hasUncommittedChanges` fails open — `git reset --hard` can destroy uncommitted changes** — `lib/commands/upgrade.dart:233-239`

```dart
final RunResult result = await _git.run(
  <String>['git', 'status', '-s'],
  workingDirectory: workingDirectory,
);
return result.stdout.trim().isNotEmpty;
```

No `throwOnError`, no exit-code check. If `git status` fails (corrupted index, permissions — stdout empty, error on stderr), the tree is reported *clean* and the flow proceeds to `attemptReset` → `git reset --hard`, destroying the user's uncommitted changes — the exact outcome this check exists to prevent. It's the only guard on that path, and a `ProcessException` here is uncaught (unlike `fetchLatestReleaseVersion`/`fetchCurrentVersion`).

Stock Flutter's `UpgradeCommandRunner.hasUncommittedChanges` fails closed: `throwOnError: true` + `on ProcessException` → `throwToolExit('The tool could not verify the status of the current flutter checkout…')`. Suggest mirroring that exactly. There's also no test for the failing-`git status` path.

---

### Important

**I1. Build never verifies the pbxproj actually links `FlutterGeneratedPluginSwiftPackage`** — `lib/build_targets/application.dart:586-638`
The umbrella is always generated, but nothing checks the Runner project references it. In a mixed state (old pbxproj + new-style Podfile that skips `Package.swift` plugins), a plugin is linked by neither system. Verification note: this is *not* silent at runtime for method-channel plugins — the registrant's `@import plugin;` fails the build — but the error is cryptic and points nowhere near the cause. The build already reads this pbxproj in `_resolveTvosDeploymentTarget` (`:643`), so when `spmPlugins.isNotEmpty` and the pbxproj lacks the umbrella reference, an actionable error is cheap. (Inverse gap: a pre-1.3 Podfile requires a podspec, so an SPM-only plugin without one is dropped on unmigrated projects.)

**I2. SPM/CocoaPods exclusivity is convention-only — double-link is reachable**
The `has_spm` skip lives only in the user-editable Podfile; nothing in `lib/` cross-checks Podfile.lock against the SPM set. `flutter_tvos` 1.1.0 itself ships **both** a podspec and a `Package.swift`, so an old Podfile (no `has_spm`) + new pbxproj links it twice — statically via the umbrella and dynamically via the pod (duplicate ObjC classes; dlsym binds one of two symbol copies). A build-time assertion (pods that also ship `Package.swift`) would make it airtight. Related: the "SPM ⇔ has Package.swift" predicate is implemented independently in three places (Dart + two Podfiles); a packaging-mode field on the plugin model would localize drift.

**I3. Umbrella assumes product name == hyphenated package name** — `lib/tvos_swift_package_manager.dart:219`
`name.replaceAll('_', '-')` matches the porter convention, but nothing ever parses a plugin's actual `.library(name:)`. A third-party plugin with any other product name fails SwiftPM resolution with an opaque error. The manifest is already regex-parsed for the package name — parsing the library name the same way removes the assumption.

**I4. `_readSwiftPackageName`: fragile parse + swallowed errors + unvalidated output** — `lib/tvos_plugins.dart:257-264`
(a) `name:\s*"([^"]+)"` with `firstMatch` over the whole file — a comment above `Package(` wins; (b) `on FileSystemException { return null; }` — no logging, silent fallback to `plugin.name`, *after* the existence check already told CocoaPods to skip the plugin; (c) the result is used unescaped as a symlink filename (`tvos_swift_package_manager.dart:102`) and interpolated into the umbrella manifest (`:164,169`). Suggest anchoring the regex to `Package(\s*name:`, logging the fallback, and validating `^[A-Za-z0-9_]+$` in `TvosSpmPlugin`. Tests only cover well-formed manifests.

**I5. Ported-plugin `Package.swift` contradicts its own doc comment on the Flutter dependency** — `lib/plugin_porting/templates.dart:162-167` vs `:205`
The doc comment says "**No Flutter dependency declared** … Declaring a Flutter package dependency here would break", yet the body emits `.package(name: "FlutterFramework", path: "../FlutterFramework")`. The design note in `tvos_swift_package_manager.dart:20-25` also says federated plugins declare no Flutter dependency. The relative path resolves inside the umbrella's `.packages/` layout (by symlink), but a standalone `swift build` / plugin CI / pub.dev analysis of a ported plugin fails unconditionally, and resolution depends on SwiftPM using the symlink path rather than the realpath — no test exercises it. One convention should win, and the comments should describe it.

**I6. FFI symbols may not survive static-archive linking — `__attribute__((used))` is not an anchor**
`flutter_tvos_ffi.h` relies on `used` + `visibility("default")`, the symbols are dlsym-only (nothing native references them), and there is no `-ObjC`/`-force_load`/`-u` anywhere in the tree (only `.linkedFramework("UIKit")`). `used` prevents dead-stripping inside *loaded* objects; it does not force selection of an unreferenced archive member, and whether Xcode links SwiftPM targets as archives or direct objects varies by version — current device verification may pass incidentally. Suggest an explicit anchor (e.g. `-u _flutter_tvos_is_tvos` in the umbrella) and an `nm`/dlsym smoke check in CI before `flutter_tvos 1.1.0` ships to pub.dev. Archive/strip (TestFlight-style) builds are untested.

---

### Downgraded after adversarial verification (for the record)

- **objectVersion 56 + Xcode-15 SPM constructs in the template** (example uses 60): real mismatch, but `xcodeproj` gem ≥1.23 and Xcode parse by ISA, not objectVersion, and stock Flutter does the same on objectVersion-54 projects. Suggest aligning to 60 for consistency; not blocking.
- **"Half-upgraded toolchain masked by *already up to date*"**: refuted — `TvosRequiredArtifacts` puts the engine into build/run/drive `requiredArtifacts`, and the stamp check against `bin/internal/engine.version` re-downloads on the next build. What remains: `throwToolExit(null, exitCode: …)` at `upgrade.dart:267-268` and `:301-302` — please give these real messages (what state the checkout is in, how to finish).
- **Removed `[CP] Embed Pods Frameworks` in the example**: refuted as a bug — with zero pods that's exactly what `pod install` produces, and it re-adds the phase when a pod appears.

### Selected minors

- Stale non-symlink entity at a link path → raw `FileSystemException` (`tvos_swift_package_manager.dart:65-69,102-106`); `fs.typeSync != notFound` → delete would fix.
- Manifests/symlinks rewritten unconditionally every build → Xcode re-resolves packages per build; compare-before-write. Stale `.packages/` symlinks of removed plugins are never pruned.
- `--verify-only` still mutates local tag refs via `git fetch --tags` (doc note); a checkout *ahead* of the latest tag is silently reset backwards.
- "Newest release" ordering is delegated to git versionsort over the full tag string (Flutter-version-primary; user `versionsort.suffix` config can reorder) — a parsed comparable tag type would be deterministic.
- Missing-xcframework path double-reports and throws a generic `Exception` instead of `throwToolExit` (`application.dart:602-606`).
- Freshly created apps (and fresh clones — `Flutter/ephemeral/` is gitignored) reference a not-yet-existing local package until the first build; consider generating the umbrella at `create` time.
- Dead test branch in `tvos_plugins_test.dart` `seedProject` (~829-836): the `widget_tvos` arm is a no-op.

Nice work overall — the atomic version pinning via a single `git reset`, the annotated-tag `^{commit}` peeling, exact SPM/Podfile predicate symmetry for freshly generated projects, and the removal of the committed `DEVELOPMENT_TEAM` all checked out clean. Engine-only Dart separation holds everywhere.

DenisovAV (approved):

PR #24, opened 2026-06-16
tvOS platform identity in AOT (release) + engine 3.44.2 (1.3.1)

MAUstaoglu:
## Summary

Ships flutter-tvos **1.3.1**: pins the engine to Flutter 3.44.2 and fixes tvOS platform identity in AOT (release/profile) builds. Bundles `flutter_tvos` plugin **1.1.1**.

## The platform-identity bug

`Platform.operatingSystem` / `isIOS` / `isTvOS` carry `@pragma("vm:platform-const")`, which `gen_snapshot` const-folds in AOT from the build's `--target-os`. Stock Flutter passes `--target-os ios` for `TargetPlatform.ios` (which tvOS rides), so release builds baked in `operatingSystem == "ios"` and `isTvOS == false` for **all** app and plugin code — debug/JIT was unaffected (no fold → reads the patched native VM). Symptom: tvOS-gated features (e.g. the Siri-Remote swipe handler) worked in debug but were dead in release.

## The fix (CLI only, no engine rebuild)

Keeps debug and release on the same runtime-resolution path:

- `TvosKernelSnapshot.build()` mirrors upstream `KernelSnapshot.build()` but passes `targetOS: null`, so the platform-const getters are not folded and resolve at runtime to `"tvos"` (the engine's patched `kHostOperatingSystemName`).
- `TvosArtifacts` overrides `flutterPatchedSdkPath` / `platformKernelDill` for AOT only to the patched `flutter_patched_sdk` in the host engine artifact, so the un-folded `isIOS` initializer is `operatingSystem == "ios" || == "tvos"` and the `isTvOS` getter exists. Debug keeps the stock SDK.

Result in both debug and release: `operatingSystem == "tvos"`, `isIOS == true`, `isTvOS == true`, `defaultTargetPlatform == iOS`. Plugin isolation is unaffected — discovery is keyed on the pubspec `tvos:` key, not `Platform.isIOS`.

## Also in this release

- **Engine → Flutter 3.44.2** (`c9a6c484`, Dart `d684a576`); all six artifact variants rebuilt (required — a version-only bump breaks AOT with `gen_snapshot: Invalid SDK hash`).
- **`flutter_tvos` 1.1.1** — the remote-control `configure` handshake now retries until the native plugin registers, so the touchpad reliably starts forwarding events instead of intermittently appearing dead.

## Tests

- New `tvos_kernel_snapshot_test.dart` asserts the AOT frontend_server command **omits** `--target-os`; `tvos_artifacts_test.dart` covers the AOT patched-SDK override + debug fall-through.
- 258 CLI + 102 plugin tests pass; `dart analyze` clean.
- Verified on a physical Apple TV in release: `tvos` / `true` / `true`, swipe works.


DenisovAV:
## Review — multi-agent pass + adversarial verification

Reviewed the AOT/kernel-snapshot logic, the artifact resolution, and the RCU change, then ran a second adversarial pass that re-verified every finding against the code at `dev`. Net: **the core fix is correct and I'd approve it** — no blockers. Two of my first-pass "critical" findings did not survive verification and are downgraded below, with the evidence.

The central mechanism checks out end-to-end (traced through the SDK + patched engine, not just the description): `Platform.operatingSystem`/`isIOS`/`isTvOS` are `@pragma("vm:platform-const")`; passing `targetOS: null` leaves them **un-folded** (the evaluator in `vm_constant_evaluator.dart:64-69,117-118` only folds when `_targetOS != null` — there's no host-OS fallback, so the "release could report `macos`" risk is firmly refuted), and the patched `flutter_patched_sdk` supplies the tvOS-aware `isIOS = (== "ios" || == "tvos")` + `isTvOS`. profile→`host_debug_unopt` / release→`host_release` is correct (both patched dills contain `isTvOS`; `profile.isPrecompiled == true`). `defaultTargetPlatform` stays `iOS`. `TvosKernelSnapshot` faithfully mirrors upstream 3.44.2 `KernelSnapshot.build()` — no real flag dropped.

One nit on the description: it says "gen_snapshot const-folds" — the fold is actually in frontend_server / the kernel transform, not gen_snapshot. The in-code comment is accurate; only the PR prose is loose.

---

### Worth fixing (non-blocking)

**I1 — RCU `_pushConfigWithRetry` gives up silently** — `packages/flutter_tvos/lib/src/rcu/tv_remote_controller.dart:250-267`
On budget exhaustion (50 × 100 ms) the loop just ends — no `reportError`, no `debugPrint`, no state flag. If the handshake genuinely never lands (a real native bug, not just the startup race), the touchpad is permanently dead with zero trace. The class already uses `FlutterError.reportError(FlutterErrorDetails(...))` two methods away (`:321`, `:368`), so this omission is internally inconsistent. Also the `catch (_)` is blanket — it swallows every type, not just the expected `MissingPluginException`/`PlatformException`, so a serialization bug in `_config.toMap()` is retried 50× then dropped indistinguishably from the race. Suggest: on exhaustion, `reportError` with the last error + attempt count; narrow the catch to the expected exceptions.

**I2 — the new RCU retry test is wall-clock-timed, not `fakeAsync`** — `packages/flutter_tvos/test/rcu/tv_remote_controller_test.dart:276-299`
Plain `test(...)` with real `await Future.delayed(250ms)` against a production 100 ms retry tick. The first assertion (empty while unregistered) is robust; the second leans on ≥1 retry tick landing inside the 250 ms window — a 150 ms margin that a starved CI runner can blow. Recommend rewriting with `fakeAsync` + `async.elapse(...)` for determinism.

**I3 — no automated end-to-end identity assertion**
`tvos_kernel_snapshot_test.dart` asserts `--target-os` is absent + exact argv; `tvos_artifacts_test.dart` asserts resolved SDK/dill paths (incl. the profile→non-product split — nicely covered). But nothing asserts the actual runtime outcome (`operatingSystem == "tvos"` / `isTvOS == true`) — that guarantee rests entirely on the manual device check in the CHANGELOG. Inherent to flag-shape testing; worth noting that a future const-fold or patched-SDK regression would pass every unit test and still break identity. A device smoke test in CI would close it (this is the same "CI against a real engine" gap we keep hitting).

### Optional hardening (not introduced by this PR)

**H1 — `_hostPatchedSdkDirectory` returns the non-nested path blindly** — `lib/tvos_artifacts.dart:110`
No `existsSync` on the fallback return. Today the override always fires for precompiled builds (verified: `TvosArtifacts` is the context `Artifacts`, and `TvosKernelSnapshot` is the only caller, always with a precompiled mode), and a genuinely-missing artifact fails *loudly* at compile (frontend-server can't find `platform_strong.dill` → throw) — so this is **not** a current silent regression, just poor diagnostics. A defensive `existsSync`-or-`throwToolExit("…run flutter-tvos precache")` would turn an opaque compile error into an actionable one, and guard against a future caller requesting the SDK with a non-precompiled mode.

**H2 — no coherence guard between `flutter.version` and `engine.version`**
A mismatch surfaces as `gen_snapshot: Invalid SDK hash`. It's late and cryptic, but it *is* surfaced — `_compileAotSnapshot` prints the raw stderr via `printError` and throws, so not silent. Both files are coherent at head, and this PR bumps them in lockstep correctly; the missing guard is pre-existing architecture, not introduced here. A cheap follow-up: assert the engine artifact's Dart revision matches the checkout (or just a comment coupling the two files).

### Minor

- `lib/build_targets/application.dart:191` — `throw Exception()` with no message on kernel-compile failure; include `errorCount` + build mode.
- The 3.44.2  remirror of `KernelSnapshot.build()` drops `_addFlavorToDartDefines`, so `appFlavor` is `null` in profile/release. No impact today (the tvOS template defines no flavors), but it's a silent divergence from the mirror — worth a one-line note/TODO.

---

Nice fix overall — the `targetOS: null` + patched-compile-SDK combination is the right call (CLI-only, no engine rebuild), the profile/release product-vs-non-product split is correct, and the artifact tests are the strongest part. My take: ship it; I'd land I1/I2 in this PR and take the rest as follow-ups.

DenisovAV (approved):

Issue #29, opened 2026-07-07
Plugin discovery does not support Dart pub workspaces


madalindk:
`flutter-tvos --version`: 1.3.1 (Flutter 3.44.2, Dart 3.12.1) macOS: 15 (Sequoia) Xcode: 16.2

**Steps to reproduce**

1. Set up a Dart workspace with multiple packages (using resolution: workspace in pubspec.yaml, as supported since Dart 3.6)
2. One of the workspace member packages is a tvOS app with plugin dependencies
3. Run dart pub get at the workspace root
4. Run flutter-tvos build tvos from the app directory

**Expected behaviour**
Plugins are discovered and registered normally.

**Actual behaviour**
The build completes with no errors, but produces a binary with no plugins registered. No warning is printed.

**Possible root cause**
`_walkPluginDependencies` in `lib/tvos_plugins.dart` reads `package_config.json` from `project.directory/.dart_tool/`:

```
final File packageConfigFile = project.directory
    .childDirectory('.dart_tool')
    .childFile('package_config.json');
```

In a Dart workspace, dart pub get only writes `package_config.json` and `package_graph.json` at the workspace root's `.dart_tool`. Member packages instead get `workspace_ref.json`, which points back to the root. Upstream Flutter's `findPackageConfig` follows this reference automatically, but `flutter-tvos` does its own file reads and doesn't know about it.

PR #32, opened 2026-07-09
fix: plugin discovery finds package_config.json in pub workspaces

thachnb-harman:
## Problem

`_walkPluginDependencies` resolves plugin paths through `<project>/.dart_tool/package_config.json` only. [Pub workspaces](https://dart.dev/tools/pub/workspaces) hoist that file to the workspace root, so for any app that is a workspace member the name→path map comes back empty and every `flutter.plugin.platforms.tvos` plugin is **silently dropped**: empty `GeneratedPluginRegistrant`, no pods or SPM packages built, and `MissingPluginException` for every plugin call at runtime. Because discovery degrades silently (by design), there is no hint that anything went wrong until the app is running.

## Fix

- Walk up from the project directory until a `.dart_tool/package_config.json` is found — the same resolution pub itself uses.
- Resolve relative `rootUri` entries against the found file's own directory (per the package_config spec) instead of assuming the project-local location. Also accept `./`-relative URIs.

Non-workspace projects are unaffected: the file is found in the project directory on the first probe, and relative `rootUri` resolution is identical to the previous behavior in that case.

## Testing

- `flutter/bin/dart analyze lib/` — no new issues.
- `flutter/bin/dart test test/` — 270 tests pass.
- Verified end-to-end on a production monorepo app (pub workspace, 7 tvOS plugins including an FFI/cargokit one): before the fix `plugins.tvos` resolved to `[]`; after, all 7 are discovered, pods/SPM packages build, and the app runs on the Apple TV simulator with plugins working.

### DIFF
--- a/lib/tvos_plugins.dart
+++ b/lib/tvos_plugins.dart
@@ -134,10 +134,21 @@ List<_DependencyPluginYaml> _walkPluginDependencies(FlutterProject project) {
       rawGraph is List<dynamic> ? rawGraph : <dynamic>[];
 
   // Build a name→path map from .dart_tool/package_config.json.
+  // Pub workspaces hoist package_config.json to the workspace root, so walk
+  // up from the project directory until one is found (the same resolution pub
+  // itself uses).
   final packagePaths = <String, String>{};
-  final File packageConfigFile = project.directory
+  File packageConfigFile = project.directory
       .childDirectory('.dart_tool')
       .childFile('package_config.json');
+  Directory packageConfigSearchDir = project.directory;
+  while (!packageConfigFile.existsSync() &&
+      packageConfigSearchDir.parent.path != packageConfigSearchDir.path) {
+    packageConfigSearchDir = packageConfigSearchDir.parent;
+    packageConfigFile = packageConfigSearchDir
+        .childDirectory('.dart_tool')
+        .childFile('package_config.json');
+  }
   if (packageConfigFile.existsSync()) {
     try {
       final packageConfig =
@@ -148,10 +159,12 @@ List<_DependencyPluginYaml> _walkPluginDependencies(FlutterProject project) {
         final pkgMap = pkg as Map<String, dynamic>;
         final name = pkgMap['name'] as String;
         var rootUri = pkgMap['rootUri'] as String;
-        // rootUri may be relative to .dart_tool/ or a file:// URI.
-        if (rootUri.startsWith('../')) {
+        // rootUri may be relative to the package_config.json's own directory
+        // (which is the workspace root's .dart_tool/ under pub workspaces) or
+        // a file:// URI.
+        if (rootUri.startsWith('../') || rootUri.startsWith('./')) {
           rootUri = globals.fs.path.normalize(
-            globals.fs.path.join(project.directory.path, '.dart_tool', rootUri),
+            globals.fs.path.join(packageConfigFile.parent.path, rootUri),
           );
         } else if (rootUri.startsWith('file://')) {
           rootUri = Uri.parse(rootUri).toFilePath();


MAUstaoglu (approved):
Reviewed against the pinned SDK (`flutter.version` = `e1fd963`) and cross-checked with the root-cause analysis in #29 — this is the right fix.

**Root cause** is exactly as described: `_walkPluginDependencies` read `package_config.json` only from `<project>/.dart_tool/`, but under pub workspaces that file is hoisted to the workspace root (members get only `workspace_ref.json`), so `packagePaths` came back empty and every `flutter.plugin.platforms.tvos` plugin was silently dropped.

**The fix is correct on both points:**

- The walk-up loop terminates properly at the filesystem root (`parent.path == path`), and finding the nearest ancestor `.dart_tool/package_config.json` is the same resolution pub's own `findPackageConfig` performs — so it lands on the real workspace root, not a false positive.
- Resolving `rootUri` against `packageConfigFile.parent.path` (the directory the file was actually found in) is spec-correct and, for the non-workspace case, byte-for-byte identical to the old `join(project.directory.path, '.dart_tool', rootUri)` — so existing projects are unaffected. Accepting `./`-relative URIs alongside `../` is a reasonable hardening.

Graceful degradation is preserved: if no `package_config.json` exists anywhere up the tree, `packageConfigFile.existsSync()` is false and `packagePaths` stays empty, same as before.

Resolves #29. Approving.

lib/tvos_plugins.dart
```diff
@@ -134,10 +134,21 @@ List<_DependencyPluginYaml> _walkPluginDependencies(FlutterProject project) {
...
   // Build a name→path map from .dart_tool/package_config.json.
+  // Pub workspaces hoist package_config.json to the workspace root, so walk
+  // up from the project directory until one is found (the same resolution pub
+  // itself uses).
   final packagePaths = <String, String>{};
-  final File packageConfigFile = project.directory
+  File packageConfigFile = project.directory
       .childDirectory('.dart_tool')
       .childFile('package_config.json');
+  Directory packageConfigSearchDir = project.directory;
```
DenisovAV, line 144: This hand-rolls `findPackageConfigFile`, which `flutter_tools` already ships (`dart/package_map.dart:25`) and which **this repo already uses** at `lib/tvos_builder.dart:79`. It's exposed as a public getter on the `FlutterProject` you already have, and `project.dart` is already imported at line 10 — so the entire block reduces to:

```dart
final File packageConfigFile = project.packageConfig;   // project.dart:227
```

That's strictly better than the loop, for two reasons beyond the dedup:

1. **It normalizes/absolutizes first** (`fileSystem.path.normalize(dir.absolute.path)`). The hand-rolled version walks `project.directory` as-is; for a relative path, `path.dirname('.') == '.'`, so `parent.path != dir.path` is false on the first check and **the walk silently doesn't walk**. In practice `project.directory` is always absolute so it never bites — but you get the guard for free.
2. **It terminates on `fileSystem.path.equals(...)`** rather than a raw `!=` (case/separator-insensitive on Windows).

It also keeps you in lockstep with upstream if pub's resolution rules ever change — and upstream's `findPlugins()` (`flutter_plugins.dart:109`), the direct analogue of this function, resolves its package config exactly this way.

Note `findPackageConfigFileOrDefault` returns a *non-existent* `<project>/.dart_tool/package_config.json` when nothing is found, so the `if (packageConfigFile.existsSync())` guard below keeps working unchanged.

lib/tvos_plugins.dart
```diff
@@ -148,10 +159,12 @@ List<_DependencyPluginYaml> _walkPluginDependencies(FlutterProject project) {
         final pkgMap = pkg as Map<String, dynamic>;
         final name = pkgMap['name'] as String;
         var rootUri = pkgMap['rootUri'] as String;
-        // rootUri may be relative to .dart_tool/ or a file:// URI.
-        if (rootUri.startsWith('../')) {
+        // rootUri may be relative to the package_config.json's own directory
```
DenisovAV, line 162: The `rootUri` resolution change itself is correct — `packageConfigFile.parent` *is* the `.dart_tool` dir, and per the package-config-v2 spec `rootUri` is resolved against the `package_config.json`'s own URI. Verified against a real workspace config.

On the `startsWith('./')` addition: I dumped three real `package_config.json` files (169 / 273 / 98 packages) and **every** non-`file://` entry starts with `../` — pub never emits `./` (it would mean the package root lives *inside* `.dart_tool/`). So the extra clause is harmless but buys nothing.

If you want to tighten this, one expression handles `../`, `./`, bare-relative, `file://`, **and** percent-decoding uniformly:

```dart
rootUri = packageConfigFile.parent.uri.resolve(rootUri).toFilePath();
```

That last part matters a little: `rootUri` is a *URI*, so a path containing a space arrives as `../my%20plugin`, and the current `../` branch does a raw string join and never decodes it (only the `file://` branch does, via `Uri.parse().toFilePath()`). Pre-existing, not introduced here — but it disappears for free if you switch.

DenisovAV (commented):
Thanks — the diagnosis is spot on and the fix is correct. I verified it empirically against a real pub workspace (members genuinely have no `.dart_tool/` of their own, and their rootUris are `../packages/<name>` relative to the workspace's `.dart_tool/`), and I confirmed there's **no regression for non-workspace projects**: when the project-local file exists the loop runs zero iterations, so `packageConfigFile.parent` is byte-identical to the old `join(project, '.dart_tool')`.

One substantive change and one gap before merge.

### 1. The upward walk already exists — and this repo already uses it

`flutter_tools` ships `findPackageConfigFile` / `findPackageConfigFileOrDefault` (`dart/package_map.dart:25,48`), exposed as `FlutterProject.packageConfig` (`project.dart:227`). Upstream's own `findPlugins()` — the direct analogue of `_walkPluginDependencies` — uses it (`flutter_plugins.dart:109`). And **this repo already calls it**: `lib/tvos_builder.dart:79` → `findPackageConfigFileOrDefault(project.directory).path`.

So the tvOS *kernel* path already resolved pub workspaces correctly while plugin discovery didn't — that asymmetry is precisely the bug. `project.packageConfig` is a public getter, `_walkPluginDependencies` already takes a `FlutterProject`, and `project.dart` is already imported here, so the whole added block collapses to one line with no new imports. Details inline.

### 2. The branch you're changing has zero test coverage

`test/general/tvos_plugins_test.dart` is ~1100 lines, but **every existing test uses an absolute `file://` rootUri** (:573, :671, :735, :778, :842, :964, :1068). The `../`-relative path this PR rewrites isn't pinned by anything today — you could break relative resolution outright and the suite stays green.

A workspace-member test would be the regression guard this fix deserves (model it on the `ensureReadyForTvosTooling end-to-end` group at :539): `/ws/.dart_tool/package_config.json` with `rootUri: '../packages/foo_tvos'`, plugin pubspec declaring `flutter.plugin.platforms.tvos`, app at `/ws/apps/my_app` with **no** local `.dart_tool` → assert `foo_tvos` reaches `GeneratedPluginRegistrant`. Fails on main, passes with this PR.

### Not a problem — the unbounded walk

In case it comes up in review: yes, the walk can in principle reach an unrelated ancestor's `.dart_tool/`, but I'd argue **against** adding a boundary. Upstream `findPackageConfigFile` has the identical unbounded walk (that *is* how pub resolves), and a pubspec boundary would stop at the workspace *member* and re-introduce the very bug you're fixing.

### Aside (not yours to fix here)

The reason this bug was undebuggable isn't the missing walk — it's `tvos_plugins.dart:171-174`, where a plugin that's in the dependency graph but unresolvable is dropped with a bare `continue`: no warning, not even a trace. The tool *knows* something is inconsistent and discards the evidence. Worth a follow-up.

PR #33, opened 2026-07-09
fix: write FLUTTER_ROOT and flutter_export_environment.sh for pod script phases

thachnb-harman:
## Problem

Upstream iOS/macOS builds write `FLUTTER_ROOT` into `Generated.xcconfig` and generate `flutter_export_environment.sh` next to it. The tvOS build does neither. Native-build tooling that runs inside CocoaPods script phases sources that file to locate the Dart SDK — [cargokit](https://github.com/irondash/cargokit) (the standard build glue for Rust-based FFI plugins) being the most common example. On tvOS its pod script phase fails with:

```
run_build_tool.sh: line 75: dart: command not found
```

because neither `FLUTTER_ROOT` nor the export script exists anywhere in the `tvos/` platform folder.

## Fix

In `_generateXcconfigs`:

- add `FLUTTER_ROOT` to `Generated.xcconfig` (from `Cache.flutterRoot`), matching upstream;
- write `tvos/Flutter/flutter_export_environment.sh` with the same variables the upstream platforms export (`FLUTTER_ROOT`, `FLUTTER_APPLICATION_PATH`, `FLUTTER_TARGET`, `FLUTTER_BUILD_DIR`, `FLUTTER_BUILD_NAME`, `FLUTTER_BUILD_NUMBER`, `COCOAPODS_PARALLEL_CODE_SIGN`).

The file is already covered by the scaffolded `tvos/.gitignore` (`Flutter/flutter_export_environment.sh`), so nothing new leaks into user VCS.

## Testing

- `flutter/bin/dart analyze lib/` — no new issues.
- `flutter/bin/dart test test/` — 270 tests pass.
- Verified end-to-end with a cargokit-based Rust FFI plugin: before the fix the pod script phase fails as above; after, cargokit locates the Dart SDK, the Rust library builds inside the Xcode build, and the app runs on the Apple TV simulator with the FFI plugin working.

### DIFF
--- a/lib/build_targets/application.dart
+++ b/lib/build_targets/application.dart
@@ -12,6 +12,7 @@ import 'package:flutter_tools/src/build_system/build_system.dart';
 import 'package:flutter_tools/src/build_system/exceptions.dart';
 import 'package:flutter_tools/src/build_system/targets/common.dart';
 import 'package:flutter_tools/src/build_system/targets/localizations.dart';
+import 'package:flutter_tools/src/cache.dart';
 import 'package:flutter_tools/src/compile.dart';
 import 'package:flutter_tools/src/dart/package_map.dart';
 import 'package:flutter_tools/src/globals.dart' as globals;
@@ -1244,7 +1245,12 @@ class NativeTvosBundle extends Target {
         ?? project.manifest.buildNumber
         ?? '1';
 
+    final String flutterRoot = globals.fs.path.normalize(
+      globals.fs.path.join(Cache.flutterRoot!),
+    );
+
     final xcconfig = StringBuffer();
+    xcconfig.writeln('FLUTTER_ROOT=$flutterRoot');
     xcconfig.writeln('FLUTTER_APPLICATION_PATH=${project.directory.path}');
     xcconfig.writeln('FLUTTER_TARGET=$targetFile');
     xcconfig.writeln('FLUTTER_BUILD_DIR=${project.directory.childDirectory('build').path}');
@@ -1252,6 +1258,24 @@ class NativeTvosBundle extends Target {
     xcconfig.writeln('FLUTTER_BUILD_NUMBER=$buildNumber');
 
     flutterDir.childFile('Generated.xcconfig').writeAsStringSync(xcconfig.toString());
+
+    // flutter_export_environment.sh — upstream iOS/macOS write this next to
+    // Generated.xcconfig. Native-build tooling that runs inside pod script
+    // phases (e.g. cargokit) sources it to locate the Dart SDK via
+    // FLUTTER_ROOT, so tvOS must provide it too.
+    final exportEnvironment = StringBuffer()
+      ..writeln('#!/bin/sh')
+      ..writeln('# This is a generated file; do not edit or check into version control.')
+      ..writeln('export "FLUTTER_ROOT=$flutterRoot"')
+      ..writeln('export "FLUTTER_APPLICATION_PATH=${project.directory.path}"')
+      ..writeln('export "FLUTTER_TARGET=$targetFile"')
+      ..writeln('export "FLUTTER_BUILD_DIR=${project.directory.childDirectory('build').path}"')
+      ..writeln('export "FLUTTER_BUILD_NAME=$buildName"')
+      ..writeln('export "FLUTTER_BUILD_NUMBER=$buildNumber"')
+      ..writeln('export "COCOAPODS_PARALLEL_CODE_SIGN=true"');
+    flutterDir
+        .childFile('flutter_export_environment.sh')
+        .writeAsStringSync(exportEnvironment.toString());
 
     // Debug.xcconfig — always write with Pods include so CocoaPods sandbox check passes.
     flutterDir


MAUstaoglu (approved):
Verified against upstream `_updateGeneratedEnvironmentVariablesScript` / `updateGeneratedXcodeProperties` in the pinned SDK — the value and format match.

- `FLUTTER_ROOT` uses `normalize(Cache.flutterRoot!)`, identical to what upstream writes into `Generated.xcconfig` for iOS. That's the value cargokit's `run_build_tool.sh` needs to find `bin/dart`.
- The `flutter_export_environment.sh` body matches the upstream script line-for-line (`#!/bin/sh`, the generated-file header, then `export "KEY=VALUE"` per variable). It's written next to `Generated.xcconfig` in `tvos/Flutter/`, matching iOS's `ios/Flutter/` layout.
- Already covered by the scaffolded `tvos/.gitignore` (`Flutter/flutter_export_environment.sh`), so nothing new leaks into user VCS. Confirmed.

Two non-blocking notes:

1. Upstream `chmod`s the export script to `755`. Here it isn't, which is fine for the sourcing path (`source`/`.` doesn't need the execute bit) and is how cargokit consumes it — but adding `globals.os.chmod(file, '755')` would match upstream and cover anything that tries to run it directly. Optional.
2. `globals.fs.path.normalize(globals.fs.path.join(Cache.flutterRoot!))` — the single-arg `join` is a no-op; `normalize(Cache.flutterRoot!)` alone is equivalent (and is exactly what upstream does). Cosmetic.

Neither blocks. Approving.

DenisovAV (commented):
Verified this end-to-end against the actual cargokit scripts, and the mechanism is right:

- `cargokit/build_pod.sh:40-52` sources `$PODS_ROOT/../Flutter/flutter_export_environment.sh`. The tvOS Podfile lives at `<app>/tvos/Podfile`, so that resolves to `<app>/tvos/Flutter/` — **exactly where you write it**. cargokit's existing iOS entry picks it up with no cargokit change.
- `run_build_tool.sh:17-21` falls back to bare `dart` when `$FLUTTER_ROOT` is unset — the reported `line 75: dart: command not found`.
- It wants `$FLUTTER_ROOT/bin/cache/dart-sdk/bin/dart` (not `bin/dart`), which exists in the vendored SDK and is *the same binary the CLI itself runs from* (`bin/internal/shared.sh:29`). So the vendored `Cache.flutterRoot` isn't merely acceptable — it's the only correct value, since a flutter-tvos user may have no stock Flutter on PATH at all.

Also confirmed the file is regenerated every build (`NativeTvosBundle` can't be skipped — `tvos_builder.dart:106-108`), and `.gitignore` already expects it (`templates/.../.gitignore.copy.tmpl:26`). `Cache.flutterRoot!` is fine too — unreachable-null (`executable.dart:95` sets it before `runner.run`), consistent with the rest of the repo *and* with upstream (`xcode_build_settings.dart:168` force-unwraps as well).

### One correction to the PR description

The `FLUTTER_ROOT=` line in `Generated.xcconfig` is **not** what fixes cargokit. `Generated.xcconfig` only reaches the **Runner** target (it's the `baseConfigurationReference` of Debug/Release.xcconfig — `project.pbxproj.tmpl:385,415,445`); the Pods-project targets, where cargokit's script phase actually runs, don't inherit it. **The `.sh` is the load-bearing half.** Keep the xcconfig line — it's correct upstream parity and useful for user-added Runner phases — just don't credit it with the fix.

### Two things to fix before merge

1. **`COCOAPODS_PARALLEL_CODE_SIGN` is in the wrong file and currently does nothing** — see the inline comment.
2. **`flutter-tvos clean` doesn't remove the new file.** `lib/commands/clean.dart:28` cleans `Flutter/Generated.xcconfig` but not its new sibling, so a stale script with an absolute `FLUTTER_ROOT` survives a clean — and someone who then moves the project (or the CLI install) and builds from Xcode directly would have cargokit source a dead SDK path. One line:

```dart
_cleanFile(tvosDir, 'Flutter/flutter_export_environment.sh');
```

### Worth adding: a test

Nothing in `test/general/` asserts on generated xcconfig content today. `NativeTvosBundle` already exposes static, MemoryFileSystem-testable builders for exactly this purpose (`copyFlutterAssetsTree`, `buildAppFrameworkInfoPlist`, `tvosGenSnapshotArgs` — see `tvos_app_bundle_test.dart`, `tvos_aot_snapshot_test.dart`). Extracting the two string builders as statics and asserting on them would pin: `FLUTTER_ROOT` present in the xcconfig; the `.sh` starts with `#!/bin/sh` and every line is `export "NAME=VALUE"`; every var in the xcconfig also appears in the `.sh` (that assertion catches the asymmetry above); and a `flutterRoot` containing a space round-trips intact.

### Checked and fine — don't change

Quoting (`export "NAME=VALUE"` is verbatim upstream and space-safe; unquoted in the xcconfig is correct, since xcconfig values are literal-to-EOL). The omitted vars (`DART_OBFUSCATION`, `TRACK_WIDGET_CREATION`, `PACKAGE_CONFIG`, …) exist upstream only to feed `xcode_backend.sh`'s in-Xcode `flutter assemble`, and the tvOS pbxproj has no such phase — flutter-tvos assembles in the CLI, so omitting them is right, not lazy. And `flutter_native_integration.env` is for SwiftPM Add-to-App (`xcode_project.dart:133`), not CocoaPods — not a gap.

lib/build_targets/application.dart
```diff
@@ -1253,6 +1259,24 @@ class NativeTvosBundle extends Target {
...
+      ..writeln('export "FLUTTER_ROOT=$flutterRoot"')
+      ..writeln('export "FLUTTER_APPLICATION_PATH=${project.directory.path}"')
+      ..writeln('export "FLUTTER_TARGET=$targetFile"')
+      ..writeln('export "FLUTTER_BUILD_DIR=${project.directory.childDirectory('build').path}"')
+      ..writeln('export "FLUTTER_BUILD_NAME=$buildName"')
+      ..writeln('export "FLUTTER_BUILD_NUMBER=$buildNumber"')
+      ..writeln('export "COCOAPODS_PARALLEL_CODE_SIGN=true"');
+    flutterDir
+        .childFile('flutter_export_environment.sh')
+        .writeAsStringSync(exportEnvironment.toString());
```
DenisovAV, line 1278: `COCOAPODS_PARALLEL_CODE_SIGN=true` only lands in the `.sh` here, which means **it currently has no effect**.

It's consumed by CocoaPods' `[CP] Embed Pods Frameworks` script phase, which reads it as an *Xcode build setting* — that phase never sources `flutter_export_environment.sh`. cargokit ignores it too. Upstream builds a single `xcodeBuildSettings` list and writes it to **both** files (`xcode_build_settings.dart:74` → xcconfig, `:114` → .sh), adding this setting at `:185` — so upstream gets it into the xcconfig, where it actually does something.

The tvOS Podfile uses `use_frameworks!` (`templates/app/swift/tvos.tmpl/Podfile:16`), so that embed phase exists here and would genuinely benefit from parallel signing. Add it to the xcconfig buffer as well (keep it in the `.sh` for upstream parity):

```dart
xcconfig.writeln('COCOAPODS_PARALLEL_CODE_SIGN=true');
```

More generally: the two buffers carrying different var sets is exactly the kind of drift a test would catch — hence the "every xcconfig var also appears in the .sh" assertion suggested in the summary.

lib/build_targets/application.dart
```diff
@@ -1244,7 +1245,12 @@ class NativeTvosBundle extends Target {
         ?? project.manifest.buildNumber
         ?? '1';
 
+    final String flutterRoot = globals.fs.path.normalize(
+      globals.fs.path.join(Cache.flutterRoot!),
```
DenisovAV, line 1249: Minor: `path.join` with a single argument is a no-op, so this reduces to `globals.fs.path.normalize(Cache.flutterRoot!)` — which is exactly what upstream writes (`xcode_build_settings.dart:168`). The `join` reads like an argument got dropped mid-edit.

Also optional, for upstream parity: upstream `chmod`s the generated script to `755` right after writing it (`xcode_build_settings.dart:124`). Not required for cargokit specifically — `build_pod.sh` *sources* the script rather than exec'ing it — but it's free (`globals` is already imported) and protects any tooling that does exec it.

PR #34, opened 2026-07-10
fix: bundle fragment shaders with the Metal runtime stage

thachnb-harman:
## Problem

`TvosCopyFlutterBundle` inherits upstream `CopyFlutterBundle.build`, which hard-codes `targetPlatform: TargetPlatform.android` for `copyAssets`. The target platform drives impellerc's runtime-stage flags (`ShaderCompiler._shaderTargetsFromTargetPlatform`): android yields SkSL/GLES/Vulkan stages, ios yields the Metal stage. Upstream can hard-code android because on real iOS the release asset copy happens in `ReleaseIosApplicationBundle`, which passes ios — but tvOS bundles assets through `CopyFlutterBundle` in every build mode, and the tvOS engine renders with Impeller/Metal.

The result: every fragment shader in a tvOS app bundle fails at load time with

```
Exception: Asset '....frag' does not contain appropriate runtime stage data for current backend (Metal).
Found stages: SkSL OpenGLES OpenGLES3 Vulkan
```

As a side effect, the build also emits impellerc SkSL-compilation warnings for shaders that can never target SkSL (e.g. ones using `fwidth` or array initializers), even though that stage is useless on tvOS.

## Fix

`TvosCopyFlutterBundle.build` now mirrors upstream `CopyFlutterBundle.build` (same maintenance contract as `TvosKernelSnapshot`: re-mirror on Flutter upgrades) with one change — `copyAssets` receives `targetPlatform: TargetPlatform.ios`, so shaders are bundled with the Metal runtime stage. The existing empty-`native_assets.json` shim is folded into the mirrored body.

## Testing

- `flutter/bin/dart analyze lib/` — no new issues vs unmodified `main`.
- `flutter/bin/dart test` — identical results with and without the change (the failures in my environment reproduce byte-for-byte on unmodified `main`; unrelated).
- End-to-end with an app using many fragment shaders: before, every `FragmentProgram.fromAsset` failed with the error above; after, the bundled `.frag` assets contain the Metal stage (MSL source present, GLES/Vulkan stages absent) and load correctly on the Apple TV simulator. The spurious SkSL impellerc warnings are gone as well.

Note: touches the same import block in `lib/build_targets/application.dart` as #33, so whichever merges second needs a trivial rebase.

### DIFF
--- a/lib/build_targets/application.dart
+++ b/lib/build_targets/application.dart
@@ -9,13 +9,18 @@ import 'package:flutter_tools/src/base/io.dart';
 import 'package:flutter_tools/src/base/logger.dart' show Status;
 import 'package:flutter_tools/src/build_info.dart';
 import 'package:flutter_tools/src/build_system/build_system.dart';
+import 'package:flutter_tools/src/build_system/depfile.dart';
 import 'package:flutter_tools/src/build_system/exceptions.dart';
+import 'package:flutter_tools/src/build_system/targets/assets.dart';
 import 'package:flutter_tools/src/build_system/targets/common.dart';
 import 'package:flutter_tools/src/build_system/targets/localizations.dart';
+import 'package:flutter_tools/src/build_system/targets/native_assets.dart';
 import 'package:flutter_tools/src/cache.dart';
 import 'package:flutter_tools/src/compile.dart';
 import 'package:flutter_tools/src/dart/package_map.dart';
+import 'package:flutter_tools/src/devfs.dart';
 import 'package:flutter_tools/src/globals.dart' as globals;
+import 'package:flutter_tools/src/isolated/native_assets/dart_hook_result.dart';
 import 'package:flutter_tools/src/project.dart';
 import 'package:meta/meta.dart';
 import 'package:package_config/package_config.dart';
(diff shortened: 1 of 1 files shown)


MAUstaoglu (approved):
Diffed the new `TvosCopyFlutterBundle.build` line-by-line against upstream `CopyFlutterBundle.build` in the pinned SDK (`e1fd963`) — it's a faithful mirror with exactly the two intended deltas:

1. the empty-`native_assets.json` shim is written first (folded in from the previous override, since we skip the native-assets targets), and
2. `copyAssets(... targetPlatform: TargetPlatform.ios ...)` instead of upstream's hard-coded `TargetPlatform.android`.

The reasoning holds: `targetPlatform` drives `ShaderCompiler._shaderTargetsFromTargetPlatform`, and only the `ios` target emits the Metal runtime stage the tvOS engine (Impeller/Metal) needs. Upstream can hard-code `android` here because on real iOS the release asset copy runs through `ReleaseIosApplicationBundle` with `ios`; tvOS routes every build mode through this target, which is why the android-stage shaders were failing at load with "does not contain appropriate runtime stage data for current backend (Metal)".

All five added imports (`depfile`, `assets`, `native_assets`, `devfs`, `dart_hook_result`) resolve in the pinned SDK and are each used (`Depfile`, `copyAssets`, `DartBuild.loadHookResult`, `DevFSContent`/`DevFSFileContent`, `DartHooksResult`).

The doc comment correctly records the maintenance contract (re-mirror on Flutter upgrades) alongside the existing `TvosKernelSnapshot` note. As flagged, this shares the import block with #33 — whichever lands second gets a trivial rebase.

Approving.

MAUstaoglu:
Approved, but this now conflicts with the base after #33 merged into `dev-fixes` — it's the shared import block in `lib/build_targets/application.dart` (your `native_assets.dart` import vs #33's `cache.dart` import). Could you rebase onto the updated `dev-fixes`? Keep both imports (alphabetical: `build_system/targets/native_assets.dart` then `cache.dart`); the rest of your change is unaffected and merges cleanly. Once it's green I'll merge.

thachnb-harman:
Rebased onto `dev-fixes` — kept both imports in the shared block, alphabetical as suggested (`build_system/targets/native_assets.dart` before `cache.dart`). The rest of the change applied cleanly; `dart analyze` is clean on the rebased branch and GitHub reports the PR mergeable again. Thanks for the thorough review!

PR #37, opened 2026-07-14
Skip plugin registration with a clear message when the engine isn't running (physical tvOS debug)

SupposedlySam:
First off, thanks for flutter-tvos — got a Flutter app building, running, and debuggable on a real Apple TV with it this week.

Sharing one rough edge that cost me a debugging session, in case a small change helps the next person.

## Why

`doc/architecture.md` notes the debug engine only starts with a debugger attached (`ptrace_check.cc`), which `flutter-tvos run` handles via its lldb/Xcode attach flow — makes sense, and it works for me. The snag is how it fails when a debug build is launched *outside* that flow: the engine never comes up, the generated registrant passes a nil registrar into the first Swift plugin, and the app SIGSEGVs with a stack that blames that plugin — nothing pointing at the real cause. I chased a phantom plugin bug for a while before landing on the ptrace note. (The `doc/debug-app.md` logging section shows a bare `devicectl … launch`, which is an easy way to hit this.)

This adds a one-time probe: if the registry hands back a nil registrar, log a clear message and skip registration instead of crashing.

## Repro

1. `flutter-tvos create demo && cd demo`, add one Swift plugin (e.g. `connectivity_plus_tvos`) so the registrant has an entry
2. set `DEVELOPMENT_TEAM`, `flutter-tvos build tvos --debug`, install on a physical Apple TV
3. `xcrun devicectl device process launch --console --device <id> <bundle id>`

## Before

```
EXC_BAD_ACCESS (SIGSEGV) at 0x0
  libswiftCore.dylib   swift_getObjectType
  Runner.debug.dylib   static ConnectivityPlusPlugin.register(with:)
  Runner.debug.dylib   +[GeneratedPluginRegistrant registerWithRegistry:]
```
(no console output — crash blames whichever plugin is first)

## After

```
[GeneratedPluginRegistrant] Flutter engine is not running; skipping plugin
registration. Debug builds on a physical Apple TV must be launched via
'flutter-tvos run' (the debug engine requires an attached debugger).
```
No crash. Verified on an Apple TV 4K (tvOS 26.5). Guard is emitted into both the ObjC and Swift registrants and only when method-channel plugins exist; `dart analyze` clean and `dart test test/` green (added two cases). A deeper fix in the engine (printing the ptrace refusal itself) would make this unnecessary — flagging in case that's preferable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@ All notable changes to flutter-tvos will be documented here.
      Do not add a version heading or bump pubspec.yaml — maintainers assign
      versions at release time. -->
 
+### Fixed
+- **Debug builds launched outside `flutter-tvos run` on a physical Apple TV no
+  longer crash during plugin registration.** The debug engine refuses to start
+  without an attached debugger, so the plugin registry hands back nil
+  registrars — which crashed the generated registrant with an unsymbolized
+  SIGSEGV blaming whichever plugin registered first. The registrant now probes
+  the registry once, logs a clear "engine is not running" message explaining
+  the `flutter-tvos run` requirement, and skips registration instead of
+  crashing. (#37)
+
 ## [1.4.0] — 2026-07-11
 
 Ships the last of the App Store submission blockers: the engine artifact is now
--- a/doc/debug-app.md
+++ b/doc/debug-app.md
@@ -103,6 +103,12 @@ xcrun devicectl device process launch --console \
 
 `<device_id>` is the UUID shown by `xcrun devicectl list devices`. The `--console` flag attaches stdout/stderr to your terminal session.
 
+> **Note:** standalone launches like this only work for **profile and release**
+> builds. A debug build launched outside `flutter-tvos run` cannot start its
+> engine (the debug engine requires an attached debugger) — plugin
+> registration is skipped and a `[GeneratedPluginRegistrant]` message
+> explaining this appears in the console.
+
 ## Other Resources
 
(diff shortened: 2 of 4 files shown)


MAUstaoglu (approved):
Thanks for this — and for the thorough writeup with the repro and before/after traces. This exact failure mode (SIGSEGV blaming the first plugin when the debug engine never started) is a genuinely nasty one to diagnose, and turning it into a one-line console message is a clear win.

Review notes:

- Verified locally: full suite green (297 tests, including your two new cases) and `dart analyze lib/` shows only the pre-existing info-level lints.
- The synthetic probe key is the right approach — probing with a real plugin key would trip the engine's duplicate-registration assert on the healthy path.
- The early return skipping the FFI forced-reference block is safe: those references do their job at link time, so runtime control flow doesn't affect symbol retention.
- Guard correctly emitted only when method-channel plugins exist, in the ObjC registrant (the one Xcode compiles).

One cosmetic nit, non-blocking: in the Swift template the guard is unconditional (emitted even for plugin-less apps), unlike the ObjC side. Since the Swift registrant isn't compiled by Xcode it has no runtime effect — fine to leave as-is.

Housekeeping done via maintainer edit: merged `dev` into the branch and added the CHANGELOG `[Unreleased]` entry (the repo just moved to a dev-based PR workflow — base was retargeted accordingly).

On your closing note: agreed the deeper fix is the engine printing its ptrace refusal — tracked for the engine repo, but this guard is worth having regardless since it also covers any future engine-init failure mode.

PR #44, opened 2026-07-25
Bump to Flutter 3.44.8 (1.4.3)

MAUstaoglu:
Pins Flutter **3.44.8** (`058e0af2c2b5`) and moves to engine artifacts `v1.0.1-flutter3.44.8`, rebuilt from source rather than republished.

The range below is quoted from 3.44.6 because the 3.44.7 artifacts were republished 3.44.6 binaries — 3.44.6 is the last version whose engine was actually compiled.

## Why this is a small change

- `DEPS` is untouched, so Dart stays at `d684a576` (a gclient `dart_revision` pin, not a submodule) and the SDK hash our AOT artifacts embed does not move — no SDK-hash gate.
- The engine-source changes are Android-side (`AccessibilityBridge.java`, `FlutterRenderer.java`, the Android toolchain GN file), an **ASan-only** copy rule in `darwin/ios/BUILD.gn`, a host-architecture fix in `copy_info_plist.py`, plus test- and CI-only files. `impeller/` is untouched.

## Correction: the size delta, and what causes it

The original description said "within ~500 bytes". That was measured from one asset and generalised. Actual range is +3 to +1204:

| asset | delta |
|---|---|
| `host_debug_unopt.zip` | +3 |
| `host_release.zip` | +4 |
| `tvos_debug_sim_arm64.zip` | +546 |
| `tvos_profile_arm64.zip` | +1191 |
| `tvos_debug_arm64.zip` | +1193 |
| `tvos_release_arm64.zip` | +1204 |

It is explained rather than build noise, which makes it a stronger argument. `copy_info_plist.py` generates the `Info.plist` embedded in `Flutter.framework` and gained `text=True` on its `subprocess.check_output` in this range, changing the recorded clang version from a bytes repr to a plain string. The two assets embedding no framework moved 3-4 bytes; all four embedding one moved 546-1204. That also corrects the original claim that none of the changed files reach a shipped `Flutter.framework` — this one does, via the `Info.plist`. No compiled code changed.

## Tooling changes that did *not* need mirroring

3.44.8 changes `UnpackDarwin.thinFramework` to run `lipo -verify_arch` once per architecture (an Xcode 27 fix). flutter-tvos never uses `UnpackDarwin` — it copies the engine framework itself — so there is nothing to mirror.

The upstream build-graph types this CLI depends on are byte-identical to 3.44.7. Correcting the original description: the relationships are not uniform, and they carry different risk on a future upgrade. `TvosKernelSnapshot`, `TvosCopyFlutterBundle` and `TvosAotElfRelease` are true subclasses and inherit upstream fixes; `TvosDartPluginRegistrantTarget` extends `Target` directly and **replaces** the stock `DartPluginRegistrantTarget`, so it inherits nothing and upstream drift there is a behavioural risk rather than a non-event. `tvosGenSnapshotArgs` mirrors `AOTSnapshotter.build` and belongs on the same checklist.

## Verification

Against the freshly built artifacts, extracted through `precache` after clearing `engine_artifacts/` so no stale extraction could be reused:

| Check | Result |
| --- | --- |
| `verify_artifacts.sh` | all pass, incl. designated-requirement origin signature + on-device JIT RWX fix |
| `dart analyze lib/` | 0 errors, 0 warnings (135 infos) |
| CLI suite | 297 passed |
| package suite | 103 passed (now wired into CI — see below) |
| tvOS 17.5 simulator | debug: launch, hot reload, hot restart, 10/10 FFI symbols, zero `Target OS is incompatible` |
| Apple TV 4K (tvOS 26.5) | debug/JIT incl. hot reload over the wireless tunnel, plus profile and release: launch + 10/10 FFI symbols |

Hot reload and restart are asserted by confirming the reloaded code actually ran, not by trusting the CLI's "Reloaded N libraries" line.

Platform identity in **every** mode: `Platform.operatingSystem == "tvos"`, `Platform.isIOS`, `Platform.isTvOS` and the package-level `FlutterTvosPlatform.isTvos` all true, with `defaultTargetPlatform == TargetPlatform.iOS`.

The 17.5 run is the one that matters for shaders — tvOS 26 tolerates a wrong metallib platform triple and older runtimes do not.

## Follow-up commit from review

CI ran `dart test test/`, path-scoped to the root suite, so `packages/flutter_tvos/test/` — 103 tests including `rcu/protocol_drift_test.dart`, which pins the RCU wire protocol against `FlutterTvRemoteProtocol.h` — never ran in either workflow. Now run with the flutter runner in both. That is the test most likely to catch engine-side drift on a bump, so it should be running on precisely this kind of PR.

Still open from review, deliberately not in this PR: a version-consistency test asserting `bin/internal/*.version` ↔ README ↔ `pubspec.yaml` ↔ CHANGELOG agree, and the stale `3.44.1` references in `doc/`.

Depends on fluttertv/engine#8. Release plumbing reminder: `flutter-tvos upgrade` is driven by git tags, so 1.4.3 is inert until `v3.44.8-tvos.1.4.3` is pushed — that must land before fluttertv/landing#6.


DenisovAV:
### Code review

Verified before commenting: tag `3.44.8` is exactly `058e0af2c2b57e369d905a03ac9748b0ebf543c6`, and the previous pin `84fc5cbb` was exactly `3.44.7` — a clean 1:1 bump. `v1.0.1-flutter3.44.8` is published with all six zips `_artifactZipNames` requires. No stale `3.44.7` or `84fc5cbb` survives anywhere outside historical changelog entries.

The core risk argument holds and I checked it directly: `DEPS` is absent from the `3.44.6...3.44.8` compare, and the only file under `packages/flutter_tools/lib/` in the range is `build_system/targets/darwin.dart` — this repo has zero references to `UnpackDarwin`, `thinFramework` or `lipo`, so nothing needed mirroring.

The "Blocked on" note at the end of the description is stale: the artifacts release published at 2026-07-25T10:53:48Z, minutes after this PR opened. Worth striking so a reviewer doesn't hold it.

**1. The "~500 bytes" figure is understated by about 2.4x.**

Measured against the published assets:

| asset | 3.44.7 | 3.44.8 | delta |
|---|---|---|---|
| `host_debug_unopt.zip` | 14 984 806 | 14 984 809 | +3 |
| `host_release.zip` | 10 893 675 | 10 893 679 | +4 |
| `tvos_debug_sim_arm64.zip` | 105 808 785 | 105 809 331 | +546 |
| `tvos_profile_arm64.zip` | 26 856 108 | 26 857 299 | +1191 |
| `tvos_debug_arm64.zip` | 106 195 959 | 106 197 152 | +1193 |
| `tvos_release_arm64.zip` | 25 388 252 | 25 389 456 | +1204 |

One zip of six is near 500 bytes. The same number appears in the engine PR and in the `v1.0.1-flutter3.44.8` release notes.

The distribution is worth stating rather than attributing it to build noise, because it is explainable. The two host zips moved 3–4 bytes; all four zips embedding `Flutter.framework` moved 546–1204. `copy_info_plist.py` — which generates the `Info.plist` shipped inside that framework — gained `text=True` on its `subprocess.check_output` in this range, changing the embedded clang version string from a bytes repr to a plain str. That also makes the claim at L19-21 slightly too strong: that change *is* visible in a shipped `Flutter.framework`. Nothing is wrong with the artifacts; an explained delta is simply a stronger argument than an unexplained one.

**2. CHANGELOG L27 misdescribes the audit surface.**

`DartPluginRegistrantTarget` is not subclassed — [`application.dart#L43`](https://github.com/fluttertv/flutter-tvos/blob/b26c1425c3b1d557522122d8255fccfd863f07dc/lib/build_targets/application.dart#L43) is `class TvosDartPluginRegistrantTarget extends Target` and *replaces* the stock target via `TvosKernelSnapshot.dependencies`. That cuts the other way from what the entry implies: a replacement does not inherit upstream fixes, so drift there is a behavioral risk rather than a non-event.

`TvosAotElfRelease` ([`L206`](https://github.com/fluttertv/flutter-tvos/blob/b26c1425c3b1d557522122d8255fccfd863f07dc/lib/build_targets/application.dart#L206)) is a genuine subclass, but `grep -rn TvosAotElfRelease lib/ test/` returns only its own declaration — it is dead in the build graph, since `ReleaseTvosApplication.dependencies` uses `TvosKernelSnapshot` + `TvosCopyFlutterBundle` and AOT runs through `NativeTvosBundle`.

The list also omits `tvosGenSnapshotArgs`, documented in-repo as mirroring `AOTSnapshotter.build` in `base/build.dart`. It is safe this time — that file is untouched in range — but it belongs on the keep-in-sync checklist.

**3. The change enumeration at L15-21 reads as exhaustive but isn't.**

The range also touches `FlutterRenderer.java` and its test, `generate_coverage.py`, `sanitizer_suppressions.sh`, `verify_exported.dart`, `.ci.yaml`, and upstream's own `bin/internal/engine.version`. None are Darwin-compiled, so the conclusion stands — but the 1.4.2 entry names `FlutterRenderer.java` as *the* 3.44.7 engine change while 1.4.3 claims to cover 3.44.6..3.44.8. A "plus test/CI-only changes" clause resolves it.

**4. The verification block records less than was actually run.**

L32-47 lists only the tvOS 17.5 simulator, while the PR body records Apple TV 4K / tvOS 26.5 across debug (including hot reload over the wireless tunnel), profile and release. Both 1.4.1 and 1.4.2 recorded on-device verification, so a reader comparing entries will conclude device testing was skipped for the one release in the series that rebuilds the engine from source. The commit message already has the right text.

---

### Coverage gaps this bump makes visible (pre-existing)

**`packages/flutter_tvos/test/` never runs in CI.** Both workflows run `flutter/bin/dart test test/` ([`test.yml#L41`](https://github.com/fluttertv/flutter-tvos/blob/b26c1425c3b1d557522122d8255fccfd863f07dc/.github/workflows/test.yml#L41), [`release.yml#L36`](https://github.com/fluttertv/flutter-tvos/blob/b26c1425c3b1d557522122d8255fccfd863f07dc/.github/workflows/release.yml#L36)), which is path-scoped to the root suite. That excludes 103 tests across 5 files — including `rcu/protocol_drift_test.dart`, which pins the RCU wire protocol against `FlutterTvRemoteProtocol.h` and is the single test most relevant to an engine bump. One-line fix in both workflows.

**Nothing asserts the two pins agree.** No test reads `bin/internal/engine.version`, `bin/internal/flutter.version`, `pubspec.yaml`'s version, README or CHANGELOG. Bump `flutter.version` and forget `engine.version` and you get real artifacts, a successful precache, a fully green suite, and a 3.44.8 kernel compiled against a 3.44.7 engine. A genuinely mistyped tag is loud, but only on a user's machine at first precache. A ~40-line offline test asserting `bin/internal/*.version` ↔ README ↔ `pubspec.yaml` ↔ CHANGELOG agree closes both cases and would also have caught the `doc/` drift five releases ago.

### Nits

- L16: Dart in the monorepo is a gclient `dart_revision` pin, not a submodule.
- L44: `isTvos` and `isTvOS` are different getters — `Platform.isTvOS` from `dart:io` versus the package's `FlutterTvosPlatform.isTvos`. The PR body distinguishes them (`isTvOS=true pkgIsTvos=true`); collapsed here it reads as a typo and drops the package-level check.
- L43: `Runner.debug.dylib` appears nowhere in this repo. The PR body's plain "10/10 FFI symbols" avoids the question.
- L15: the range starting at 3.44.6 rather than 3.44.7 reads as a typo without half a sentence noting that the 3.44.7 artifacts were republished 3.44.6 binaries.
- `doc/architecture.md` L5 and L199, and `doc/get-started.md` L54, still say Flutter 3.44.1 — stale since 3.44.2. Since these display a version, they belong in the same sweep; better still, point them at `bin/internal/flutter.version` so they stop drifting.
- Release plumbing reminder: `flutter-tvos upgrade` is driven purely by git tags, and nothing reads `pubspec.yaml` at runtime — so 1.4.3 is inert until `v3.44.8-tvos.1.4.3` is pushed.

DenisovAV (approved):
Approving. Pins verified: tag 3.44.8 is exactly `058e0af2c2b57e369d905a03ac9748b0ebf543c6`, the previous pin was exactly 3.44.7, and `v1.0.1-flutter3.44.8` is published with all six zips `_artifactZipNames` requires. The no-risk argument holds — `DEPS` is absent from the `3.44.6...3.44.8` compare and the only `flutter_tools/lib/` file in range is `targets/darwin.dart`, which this repo never touches.

The "Blocked on" note in the description is stale; the artifacts release published minutes after this PR opened.

Non-blocking, from the detailed review above: the "~500 bytes" figure is understated (actual max +1204, and the delta is explainable via `copy_info_plist.py`), the subclass list needs a correction, and the changelog records less verification than was actually run. Separately worth doing soon: `packages/flutter_tvos/test/` never runs in CI, which excludes the one test written to catch engine-side drift on exactly this kind of bump.

Merge after fluttertv/engine#8, and push the `v3.44.8-tvos.1.4.3` tag before fluttertv/landing#6 goes out.

PR #50, opened 2026-08-02
Version selection: flutter-tvos versions / use <version>

DenisovAV:
Adds `flutter-tvos versions` and `flutter-tvos use <version>`, so a user can pick which Flutter version this checkout targets.

**This ships the mechanism, not a new supported version.** A version appears in the list when a `v<flutter>-tvos.<tool>` tag exists for it. Support for 3.32.8 additionally needs the engine patch set (fluttertv/engine#11), signed engine artifacts, and a CLI port to 3.32.8's `flutter_tools` API — none of which this PR contains.

### Why release lines

The CLI imports ~75 **private** `package:flutter_tools/src/...` libraries, which upstream changes without notice. Measured against a real 3.32.8 checkout, `main`'s source produces 21 analyzer errors — `BuildCommand` takes 6 constructor parameters there and 19 at 3.44, `src/ios/lldb.dart` does not exist before 3.35, `Device.isSupported` changed from `bool` to `Future<bool>`.

So one source tree cannot span versions. Each supported version becomes a release line: the tag carries the CLI ported to that version's API *and* both pin files. Switching is therefore a checkout of a tag — which is also what Flutter itself tells you to do (`flutter downgrade` refuses a version argument and links to a doc that says `git checkout <tag>`). flutter-tizen, whose `shared.sh` ours mirrors, uses the same tag scheme and never automated the switch at all. This automates what upstream leaves manual.

Version discovery reads git tags rather than a manifest or the Releases API: the tag is already the operative artifact, so any other source is a claim *about* tags that can drift from them. It also degrades best offline — previously fetched tags stay listable and checkout-able.

### Commands

```
flutter-tvos versions          # one line per Flutter version, current marked; --all for every tag
flutter-tvos use <version>     # 3.44.5, or an exact tag; --force to switch past a dirty tree
flutter-tvos upgrade           # unchanged
```

`channel` is **unregistered** and `downgrade` is **removed**. Both were forwarded stock, and both set `workingDirectory = Cache.flutterRoot` — the vendored SDK — then reset hard in it, breaking the `flutter.version` ↔ engine-artifact pin. That is the exact hazard `TvosUpgradeCommand` exists to avoid. `use <version>` covers what `downgrade` offered, explicitly.

### Review

Eight agents. Four criticals found, each reproduced before being believed, all fixed in-branch:

- **`git reset --hard` destroyed committed-but-unpushed work.** The dirty-tree guard reads `git status -s`, which reports the worktree and is silent about the branch pointer, so a contributor on `main` with local commits passed it cleanly and lost them to the reflog. Now `git checkout --force --detach` — identical worktree, no ref touched. Applied to `upgrade` too, where the cost had been documented as accepted.
- **The recovery command was word-wrapped into a command that lies.** It went through `throwToolExit` → `printError`, which wraps to terminal width; a wrapped `git … --detach <tag>` pastes as a checkout with no revision, which succeeds against HEAD, exits 0, and prints a reassuring message while the user is still stranded. Now printed with `wrap: false`, and unconditionally the moment the checkout moves.
- **`downgrade <version>` silently ignored the argument** while inheriting `--force`, so it discarded uncommitted work and switched to the recorded tag rather than the one typed. Moot now that the command is gone, but it is why it went.
- **A failing `git tag -l` had no catch**, so an ordinary git refusal became "Oops; flutter has exited unexpectedly" plus a crash report.

Mutation testing found the suite blind to the one invariant the feature protects: no stub pinned a working directory, so redirecting `git reset --hard` *into the vendored SDK* passed every test. 28 stubs now pin it. The real-git sort test was likewise toothless — every component was single-digit, so it passed under a sort with no version semantics at all, which is precisely what it was written to rule out.

Full report: `test_reports/pr-reviews/pr-branch-feat-cli-version-selection-review.md`.

### Verification

337 tests pass. Analyzer at `dev`'s own 191-issue baseline, none in new code. Commands exercised live.

### Known follow-ups

Not blocking, listed in the report: `resolve()` states "no release matches" as fact when the tag list is known stale; `versions`/`use` do not set `shouldUpdateCache => false`, so they force an artifact update before running; and `versions` does not override `invocation`, so its help line reads `Usage: flutter versions`.


MAUstaoglu (changes requested):
Good work. `reset --hard` -> `checkout --force --detach` is the right call, and the mutation testing finding is a nice catch — a suite that passes while the reset goes into the vendored SDK is exactly the blind spot worth hunting.

One thing to fix: `use.dart:136` tells the user to go back with `flutter-tvos downgrade`, but this PR removes that command. It's on the precache failure path, so it's the wrong instruction at the worst moment. Point it at `use <version>` or the git line you already print above it.

Two more I'd like before merge:

`versions` doesn't override `shouldUpdateCache`, so listing git tags kicks off an engine artifact download first. On a fresh clone that's a few hundred MB to print a list of tags. One line, and `use` has the same thing but sillier — it updates the cache for the version you're leaving.

The recovery line doesn't get people back on their branch. `current()` only keeps hash and tag, so someone who was on `main` with local commits gets a detached checkout at the right commit. Their commits are safe now, which was the point, but they're not back where they were. Reading `git symbolic-ref -q --short HEAD` and preferring it would finish the fix.

Last one is your call, not a defect: this commits ~2400 lines of plans, specs and review reports, which is more than half the diff. Do we want those in the repo permanently, or somewhere else?

DenisovAV:
All three fixed, pushed.

**The `downgrade` reference was mine and embarrassing** — I wrote that message while the command still existed and never revisited it when it went, so the worst moment carried the one instruction guaranteed to fail. It now points at the git line already on screen.

**`shouldUpdateCache`** is set on both. You were right that `use` is the sillier one: it refreshed the cache for the version it was about to leave, which `shared.sh` then deletes along with `bin/cache` on the switch.

**The branch restoration is the best catch of the three.** "Their commits are safe now, but they're not back where they were" is exactly it — I'd fixed the data loss and stopped, treating *safe* and *back* as the same thing. `current()` now reads `git symbolic-ref -q --short HEAD` and the recovery line prefers it, falling back to the detaching form only when HEAD was already detached. Pinned by a test, since the case that matters is the untagged development checkout a fake would otherwise never exercise.

On the docs: dropped the implementation plan (2068 lines of scaffolding that had done its job). Kept the design spec and the review report — the first explains why release lines are shaped this way, which is the thing a future reader will otherwise have to reconstruct; the second records the hazards found and the follow-ups still open. Docs are now ~480 lines against 1750 of code.

Worth flagging separately, since it affects how this gets used: **no existing release tag carries `versions`/`use`.** They ship in the next one. So `use` is currently one-way into the past — you can switch back to 3.41.9, but once there the command is gone and the printed git line is the only way forward again. Either we backport into each live line and re-tag, or we document the asymmetry. Happy to do either.

DenisovAV:
Correction to my previous comment: I said docs were "~480 lines" — they were 3532 across 20 files at that moment. An over-broad `git add -A docs` had swept in eighteen files sitting untracked from earlier sessions: a build guide, customer correspondence, a `.gclient`, and a directory of PR reviews for unrelated changes. None of it was reviewed here and some of it is not what you want landing in a public repo by accident.

Removed, and they stay untracked as before. The docs in this PR are now genuinely 483 lines — the design spec and this PR's own review report.

DenisovAV:
Dropped the spec and the review report as well — this PR is now code and tests only, 14 files.

That keeps it consistent with everything else here: a dozen reviews written for this project live on disk and none of them is in git, so committing this one would have made it the odd one out rather than starting a practice. The reasoning that matters is in the description and in the code comments, which travel with the code.

Also folded the rule into the review skill so the next run does not re-litigate it, along with a note not to stage `docs/` wholesale in this repo — it is public, `docs/` is tracked, and the untracked files sitting next to it are not all things to publish.

DenisovAV:
Added the `docs/licensing/` ignore here (was #51, now closed).

This repo is public and `docs/` is tracked, so that untracked licensing package — source licence, CLA, NDA, the § 8 UrhG co-authors agreement — sits one `git add docs/` away from being published, and that is not undoable once GitHub has cached it. Nothing from it is in history; the line only closes the accident.

Same root cause as the eighteen unrelated files I had to back out of this PR earlier, so it seemed worth fixing rather than remembering.

MAUstaoglu (approved):
All three are addressed, and the branch restoration went further than I asked — carrying the branch on `TvosVersion` and preferring it over the tag is the right shape, not just a patched-up message.

Good catch on the eighteen files. Glad that surfaced here rather than after a merge.

LGTM.

PR #51, opened 2026-08-03
Ignore docs/licensing/

DenisovAV:
One line. This repo is public and `docs/` is tracked, so the licensing package sitting untracked inside it — source licence, CLA, NDA, the § 8 UrhG co-authors agreement, a risks-accepted memo — is one `git add docs/` away from being published, and that cannot be undone once GitHub has cached it and forks have diverged.

Nothing from it is in history. This only closes the accident.

Found while preparing #50, where a broad `git add -A docs` swept in eighteen unrelated untracked files and I had to back them out. Same root cause, so worth fixing rather than remembering.


DenisovAV:
Folded into #50 — same one-line change, one less PR to shepherd.

PR #54, opened 2026-08-04
Fix `flutter-tvos create .` naming the project "."

MAUstaoglu:
## Problem

```bash
flutter-tvos create --platforms=tvos .
```

scaffolds a project literally named `.` — `name: .` in `pubspec.yaml`, a `.App`
widget class in `lib/main.dart`, and a bundle identifier built from `.`. The
result doesn't `pub get`, let alone build.

## Cause

`TvosCreateCommand.runCommand` derived both values from the raw argument:

```dart
final String projectDirPath = argResults!.rest.first;                  // "."
final String name = stringArg('project-name')
    ?? globals.fs.path.basename(projectDirPath);                       // "."
```

`basename(".")` is `"."`. Stock `flutter create` avoids this because
`CreateBase` already exposes a `projectDirPath` getter (normalized + absolute)
and a `projectName` getter with the full resolution chain and name validation —
this command just wasn't using them.

## Fix

Use the inherited getters. Behaviour now matches stock `flutter create`:

- `.` resolves to the containing directory's name.
- Re-running inside an existing project reuses that project's `pubspec.yaml`
  `name` instead of the folder name.
- `--project-name` still takes precedence.
- An illegal package name (e.g. a `my-tv-app` directory) exits with
  *"my-tv-app" is not a valid Dart package name* instead of writing a broken
  project.

## Tests

New `test/general/tvos_create_test.dart` covers all four cases. Verified they
fail against the previous code (`Actual: 'name: .\n'`) and pass with the fix.
Full suite: 342 passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to flutter-tvos will be documented here.
 
 ## [Unreleased]
 
+### Fixed
+
+- `flutter-tvos create --platforms=tvos .` no longer names the project `.`.
+  The command derived the package name from the raw output-directory argument,
+  so a `.` target produced `name: .` in `pubspec.yaml` (and a `.App` widget
+  class). It now uses the same resolution as stock `flutter create`:
+  `--project-name` if given, else the `name` of an existing `pubspec.yaml`,
+  else the basename of the *normalized absolute* directory path — and the
+  result is validated, so a directory like `my-tv-app` fails with
+  "not a valid Dart package name" instead of scaffolding a broken project.
+
 ## [1.4.4] - 2026-07-31
 
 - Engine artifacts updated to `v1.0.2-flutter3.44.8` (origin-signed for ITMS-91065
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,14 @@ All notable changes to flutter-tvos will be documented here.
   result is validated, so a directory like `my-tv-app` fails with
   "not a valid Dart package name" instead of scaffolding a broken project.
 
+  Without `--platforms=tvos` the damage was quieter but still there: upstream
+  `flutter create` resolved the Dart package name correctly, while the `tvos/`
+  runner we add on top got `CFBundleName = .` and a bundle identifier that
+  collapsed to a bare `com.example` — shared by every project created that way.
+  If you generated a project with `create .`, check
+  `tvos/Runner/Info.plist` and `PRODUCT_BUNDLE_IDENTIFIER` in
+  `tvos/Runner.xcodeproj/project.pbxproj`.
+
 ## [1.4.4] - 2026-07-31
 
 - Engine artifacts updated to `v1.0.2-flutter3.44.8` (origin-signed for ITMS-91065
--- a/lib/commands/create.dart
+++ b/lib/commands/create.dart
@@ -30,9 +30,15 @@ class TvosCreateCommand extends CreateCommand {
     // below reads `rest.first` before delegating to `super.runCommand()`, so
     // validate up front.
     validateOutputDirectoryArg();
-    final String projectDirPath = argResults!.rest.first;
-    final String name =
-        stringArg('project-name') ?? globals.fs.path.basename(projectDirPath);
+    // Use the inherited `CreateBase` getters rather than reading
+    // `argResults!.rest.first` raw: `projectDirPath` is normalized and
+    // absolute (so `flutter-tvos create .` resolves to the current directory
+    // instead of yielding the literal "."), and `projectName` mirrors stock
+    // `flutter create` — explicit `--project-name`, else the `name` of an
+    // existing pubspec.yaml, else the directory basename — and validates the
+    // result is a legal Dart package name.
+    final String projectPath = projectDirPath;
+    final String name = projectName;
     final String templateType = stringArg('template') ?? 'app';
 
     // tvOS-only app: build the shared scaffold + tvos/ ourselves. We do
@@ -41,8 +47,8 @@ class TvosCreateCommand extends CreateCommand {
     // then stripped — the project is tvOS-only by construction.
     if (boolArg('tvos-only') && templateType != 'plugin') {
       globals.logger.printStatus('Generating tvOS-only project...');
-      TvosAppScaffold(globals.fs).write(projectDirPath, name);
-      await _renderTvosRunner(projectDirPath, name);
+      TvosAppScaffold(globals.fs).write(projectPath, name);
+      await _renderTvosRunner(projectPath, name);
       globals.logger.printStatus(
         'Created tvOS-only project (shared app + tvos/, no other platforms).',
       );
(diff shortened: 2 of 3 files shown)


DenisovAV (approved):
Approving. The fix is the right one — using the inherited getters instead of
re-deriving what `CreateBase` already resolves — and I checked the claims rather
than taking the description's word for them.

**The three claims about `CreateBase` hold**, against the vendored tree:
`projectDirPath` is `normalize(projectDir.absolute.path)` (`create_base.dart:69`),
and `projectName` (`:269`) really is `--project-name` → the `name` of an existing
`pubspec.yaml` → the directory basename, validated at `:298` unless
`--skip-name-checks`. That last one was worth confirming for a second reason:
`boolArg('skip-name-checks')` would throw if the option were not registered, and
it is — `create.dart:89`, inherited.

**The tests do catch the bug.** I reverted `lib/commands/create.dart` to
`520a54b` and re-ran them: the `create .` case and the invalid-name case fail
(`Expected: contains 'name: my_tv_app'` / `emitted <null>`), the other two pass.
So two of the four are genuine bug-catchers and two are regression guards —
worth knowing which is which, not a complaint. Full suite green at 342, matching
your number.

**No behaviour regression from validating earlier.** Reading `projectName` at the
top now throws before anything is written, where previously the standard path
reached the same `throwToolExit` inside `super.runCommand()`. Same error, just
before files land instead of after.

Two notes, neither blocking.

**The standard path is still uncovered, and it is affected.** All four tests pass
`--tvos-only`, which returns before `super.runCommand()`. But `flutter-tvos
create .` with no `--platforms` takes the other branch, and the same `name` flows
into `_renderTvosRunner` → `'projectName': name` and
`createUTIIdentifier(organization, name)` (`tvos_runner.dart:45,56`). So pre-fix
that invocation produced a *correctly* named Dart project with a `tvos/` runner
named `.` and a bundle identifier built from `.` — split-brain rather than
uniformly broken, which is the harder version to notice.

I tried to add a probe for it and stopped: the standard path needs the real
template tree and dies on `FileSystemException: ... path = 'flutter/packages'`
under `MemoryFileSystem`. So it is not a cheap test to add, which I assume is why
you scoped to `--tvos-only`. Fine to leave; flagging it so the gap is on record
rather than implied to be covered.

**One trap for whoever writes the next create test.** My first probe used
`--platforms=tvos` and failed with *"tvos" is not an allowed value* — because the
rewrite to `--tvos-only` lives in `expandTvosPlatformArgs` at
`executable.dart:197`, and `createTestCommandRunner` goes straight to the command
and never sees it. Command-level tests bypass the shim, so `--platforms=tvos` has
to be spelled `--tvos-only` in tests. That cost me a wrong conclusion for a few
minutes; worth a line in the test file at some point.

DenisovAV:
Thanks for writing the standard-path test — and the shim note, which is exactly
the trap I fell into.

**The new test is red in CI and green locally, and it can only ever be green
locally.** Worth fixing before merge, because as it stands it is a test that
passes on developer machines for a reason unrelated to the code it covers.

`renderMerged` → `Template.merged` → `TemplatePathProvider.imageDirectory`
resolves `flutter_template_images` out of

```
$Cache.flutterRoot/packages/flutter_tools/.dart_tool/package_config.json
```

(`template.dart:47-51`). That file only exists once something has run pub get
*inside the vendored `flutter_tools`* — which `bin/flutter-tvos` does via
`shared.sh` when it builds the tool snapshot. On my machine it is there, dated
the last time I bootstrapped, so the test passes.

CI never bootstraps. `test.yml` clones Flutter at `--depth=1` and its
"Resolve dependencies" step is `flutter/bin/dart pub get` at the **repo root**,
which writes `./.dart_tool/package_config.json` — the flutter-tvos package, not
`flutter_tools`. So the path above is absent, `loadPackageConfigWithLogging`
errors and `throwToolExit('')` fires, which is why the log shows a bare
`Error:` with no message above the stack.

Cheapest fix is one more step before "Run unit tests":

```yaml
- name: Resolve flutter_tools dependencies
  run: flutter/bin/dart pub get --directory flutter/packages/flutter_tools
```

It has to be online, so it goes before the `PUB_OFFLINE: "1"` step rather than
inside it. I have not run that against a fresh `--depth=1` clone — my checkout is
already resolved, so I cannot honestly tell you it succeeds there; worth
confirming on the CI run rather than taking my word for it.

The alternative is to have the test skip when the package config is missing, and
I would argue against it: a create test that quietly no-ops on the one machine
that gates merges is worse than not having it.

Nothing else changed in my read — approval stands, this is the CI wiring rather
than the fix.

MAUstaoglu:
Your comment raced the fix by about 85 seconds — `ec81142` landed at 13:30, the
comment is timestamped 13:31, so you were reading `362757f` and the red run. CI
is green on `ec81142` now (343 CLI + 103 package tests), and the log shows all
five create tests executing rather than skipping.

We diagnosed it identically, including the bare `Error:` being `throwToolExit('')`
out of `loadPackageConfigWithLogging`, and `bin/flutter-tvos` → `shared.sh` being
what warms `flutter_tools` on a dev machine. Your read of `test.yml` is right too:
the root-level `pub get` writes the flutter-tvos package config, never
`flutter_tools`'.

I went with a different remedy: override the injectable `TemplatePathProvider` in
the test so it resolves `flutter_template_images` from *our*
`.dart_tool/package_config.json`, which `dart pub get` always produces — the suite
cannot run without it. CI wiring is untouched.

I preferred that over the extra CI step for two reasons. It adds no online pub
resolution to a workflow that sets `PUB_OFFLINE=1` specifically because of pub
flakiness. And there is a risk in `dart pub get --directory
flutter/packages/flutter_tools` beyond the one you flagged: it does version
solving against the SDK's own pins, and the `--depth=1 --tags` handling in
`test.yml` exists because a shallow clone otherwise reports `0.0.0-unknown` —
exactly the thing that breaks SDK-constrained resolution. The override sidesteps
that question instead of betting on it.

To your point about not taking anyone's word for it: I verified by hiding
`flutter/packages/flutter_tools/.dart_tool` to reproduce a cold checkout. The
previous version of the test fails there with the same stack CI produced; this
one passes; and it still fails against the pre-fix command, so it has not been
defanged into passing for the wrong reason.

Agreed on rejecting the skip-when-missing option, for the reason you gave.

One cost worth naming: the test no longer exercises upstream's real
image-directory resolution — the override replicates that logic against a
different package config. For a test about naming I think that is the right
trade, but it is a fidelity loss rather than a free win. If you would rather have
the CI step so the real path stays covered, I will verify it against a fresh
`--depth=1` clone before pushing it.

DenisovAV:
You beat me to it, and with the better fix — my comment above was written
against `362757f` and proposed adding a `pub get` step to `test.yml`. Overriding
`TemplatePathProvider` is right and mine was not: a CI step would have satisfied
the dependency, yours removes it, so the test stops caring whether the SDK has
been bootstrapped at all rather than needing CI to bootstrap it. Disregard that
suggestion.

Confirmed your cold-checkout verification independently: renamed
`flutter/packages/flutter_tools/.dart_tool` out of the way and ran the file —
all five pass. Restored it after.

Approval stands, and CI is green now.

PR #55, opened 2026-08-05
Sync 1.5.0 release commit back into dev

MAUstaoglu:
`main` was released as 1.5.0 (`c63b07e`, tag `v3.44.8-tvos.1.5.0`) but `dev` never received the release-prep commit — `dev` requires a PR with one review, so it could not be pushed directly at release time.

The content difference is exactly three files:

- `CHANGELOG.md` — the `## [1.5.0]` heading and its entries (currently still under `[Unreleased]` on dev)
- `README.md` — the current-version block
- `pubspec.yaml` — `1.4.3` → `1.5.0`

Everything else in dev's seven commits is already in main's squash, so this carries no code changes.

Worth merging before the next release rather than letting it accumulate: main's squash commit is not in dev's history, and that divergence is what forced a squash this time round instead of a fast-forward.


MAUstaoglu:
Superseded by a branch carrying only the three version files — this PR diffed from the merge base and showed all 27 files as conflicting, an artifact of main having squashed dev's commits.

PR #62, opened 2026-08-12
Derive the generated registrant's @dart marker from the SDK

MAUstaoglu:
`writeTvosDartPluginRegistrant` hardcoded `// @dart = 3.9` into the generated `.dart_tool/flutter_build/dart_plugin_registrant.dart`.

That is latent on Flutter 3.44.x, whose Dart accepts 3.9, but it breaks the kernel compile outright on any Flutter shipping an older Dart:

```
The specified language version 3.9 is too high.
The highest supported language version is 3.8.
```

Found while porting the CLI to Flutter 3.32.8 (Dart 3.8.1) — the `release/flutter-3.32.8` line hits this immediately.

## Fix

Derive the marker from the SDK actually in use:

```dart
final LanguageVersion languageVersion = currentLanguageVersion(globals.fs, Cache.flutterRoot!);
```

`currentLanguageVersion` is the same helper upstream's own tooling resolves to, and it lives at the same path (`flutter_tools/src/dart/language_version.dart`) in both 3.32.8 and 3.44.x — so the change is portable across every Flutter line this CLI targets, rather than trading one hardcoded value for another.

## Why it matters beyond the 3.32.8 port

The generated registrant is what makes federated `*_tvos` Dart plugins register at isolate startup. A too-high language marker fails the kernel compile before the app ever builds, so the symptom is a hard build failure on older Flutter lines rather than a silent degradation.

## Test

Adds a regression test that pins the marker to the derived value and explicitly rejects the old literal, so a future hardcode reintroducing `// @dart = 3.9` fails the suite.

## Validation

- `tvos_plugins_test.dart` — 42 passed, including the new case
- `dart analyze lib/` — 0 errors, 0 warnings


MAUstaoglu:
Superseded by #63.

This branch is the base of `chore/flutter-3.47.0`, so `ce00747` ships there unchanged — closing rather than merging deliberately: a squash-merge here would replay the same change and conflict with #63 on `CHANGELOG.md`.

The registrant fix was in place throughout #63's verification (349 unit tests, plus build/run on a tvOS simulator and a physical Apple TV).

PR #63, opened 2026-08-13
Flutter 3.47.0 on the new tvOS engine, and fix on-device debug hot reload

MAUstaoglu:
Moves the CLI to Flutter 3.47.0 against the matching tvOS engine, fixes on-device debug, releases **1.6.0**, and replaces the `dev` → `main` pull request with an automatic mirror.

**Supersedes #62** — its commit (`ce00747`, the registrant `@dart` marker fix) is the base of this branch, so it ships here. #62 is closed rather than merged deliberately: squash-merging it first would replay the same change and conflict on `CHANGELOG.md`.

## Engine artifacts

`bin/internal/engine.version` pins `engine-495915c579071a4ef6a5014637166d5231cb4aac`,
which is published in `fluttertv/engine-artifacts` with all six variants. Verified by
hiding any local artifacts and running `precache` against GitHub: all six downloaded,
every artifact check passed (platform triples, on-device JIT fix, platform-identity
patch, origin signing), and the compile SDK reads the expected kernel version.

The tag is prefixed rather than a bare commit SHA because GitHub rejects any ref that
is exactly 40 or 64 hex characters as ambiguous with a commit id.

## Flutter 3.47.0

Pin moves to `4cf24164` (`flutter-3.47-candidate.0`). Three upstream API breaks, all in the code that deliberately shadows `flutter_tools`' build graph:

- **`DartBuild` → `LinkHooks`** — `TvosCopyFlutterBundle` mirrors upstream's `build()`, which loads the hook result. Rename only.
- **`LLDB` now requires an `XcodeProjectInterpreter`.**
- **`KernelSnapshot` gained a `recorded_uses.json` output**, declared whenever the record-use experiment is on — which is by default, on every channel. `TvosKernelSnapshot` inherits that output list, so its mirrored `build()` has to produce the file (`--recorded-uses` in AOT, an empty file in JIT) or the build fails on an output nothing wrote. Covered both ways in `tvos_kernel_snapshot_test.dart`.

None of this surfaces until a real build — exactly the re-mirroring hazard the upgrade notes warn about.

## On-device debug: hot reload was silently dead

`flutter-tvos run -d <appletv> --debug` launches with `--vm-service-host=0.0.0.0`, so the URI the Dart VM prints has a host the Mac cannot dial; it has to be rewritten to the device's LAN address. That rewrite got a **single 10 s mDNS query**, and when it lost the race with Bonjour propagation it returned the `0.0.0.0` URI anyway. The resident runner accepts it and retries forever:

```
SocketException: Connection refused, address = 0.0.0.0, port = 59104
```

The session looks alive — the app on the TV is completely healthy — while hot reload, hot restart and DevTools do nothing, and nothing says why.

`devicectl` cannot rescue it: a LAN-paired Apple TV reports `tunnelState: "disconnected"` with no `localHostnames` and no `ipAddress`, so `_resolveDeviceIp` returns null every time. Bonjour is the only real source.

Now: poll mDNS in 5 s attempts up to 60 s (`FLUTTER_TVOS_MDNS_TIMEOUT_SECONDS`), keep the port match so a stale copy of the same bundle id still running on the TV can't win the lookup, don't let a first-attempt missing-permission `ToolExit` abort a run that later attempts would rescue, and when nothing resolves return **no** URI plus a warning naming Local Network permission — rather than one already known to be dead.

## Test suite on 3.47.0

31 tests started failing for a reason that isn't ours. The new FS guard (`test/src/fs_safety.dart`) rejects writes outside the system temp directory, but resolves symlinks only when computing the allowed root — not on the path it checks. On macOS `$TMPDIR` is `/var/folders/…` behind a `/var` → `/private/var` symlink, so the two never compare equal and anything touching a real temp directory dies, including code paths that only reach it through `flutter_tools`' own `LocalFileSystem`.

Fixed by running with an already-resolved `TMPDIR`, which removes the mismatch at the source and leaves the guard armed — not with `FLUTTER_TEST_DISABLE_FS_GUARD`, which would switch off the thing that stops a stray test writing to `$HOME`. Applied in both workflows and documented in `test/README.md` for local runs.

The other 2 failures were ours: `TvosKernelSnapshot.build` now reads `featureFlags`, a context lookup, so those tests need a context.

## Verification

- `dart analyze lib/` — 0 errors, 0 warnings.
- Unit suite — **349 passing**.
- Bundled package suite — unchanged.

Built and run against the **published** engine artifacts — downloaded from the release, not built locally — on a **tvOS 17.5 simulator** and a **physical Apple TV 4K**, across debug, profile and release:

| | simulator | device |
|---|---|---|
| launch + VM Service | ✅ | ✅ |
| hot reload / hot restart (verified *effective*, not just reported) | ✅ | ✅ |
| platform identity | ✅ | ✅ |
| FFI symbol export | ✅ | ✅ |

`device-debug` was additionally repeated 4× with the app killed on the device between runs, so each launch had to re-publish its Bonjour record — green every time, and green again on the final run against the published artifacts.

Now runtime-verified rather than assumed:

- **Platform identity in every mode** — `os=tvos isIOS=true isTvOS=true pkgIsTvos=true target=TargetPlatform.iOS ffiIsTvOS=true`, on simulator and hardware, in debug, profile and release. The `targetOS: null` + patched-compile-SDK path holds on 3.47.0.
- **Real hardware** — `ffiMachine=AppleTV14,1 ffiSimulator=false`, with the on-device JIT RWX path and Impeller/Metal both active in the device log.
- **FFI linking** — 10 `_flutter_tvos_*` symbols exported in all four app binaries, including the AOT `Runner`.

## Release 1.6.0

`pubspec.yaml`, the README "Current version" block and the `CHANGELOG` heading all move together — the README one is the easy miss, and it would otherwise keep advertising 3.44.9.

Minor rather than patch: 1.5.1 was a same-line SDK bump (3.44.8 → 3.44.9, identical Dart revision, no engine rebuild). This is a new engine, three re-mirrored `flutter_tools` APIs, and a fix for a feature that was silently non-functional on hardware. The pending `[Unreleased]` entry (the registrant `@dart` marker, originally #62) is folded into `[1.6.0]` rather than left dangling.

The bundled `packages/flutter_tvos` is deliberately **not** bumped — nothing in this diff touches its source.

## Review round (`fdc2ba4`)

A review of this branch turned up 12 findings; all are fixed in `fdc2ba4`. The two that mattered:

**The mDNS retry loop had no backoff, and the case it was written for fails fast.** With `throwOnMissingLocalNetworkPermissionsError: false`, upstream catches the macOS socket error and returns an empty result in *milliseconds* rather than consuming the attempt timeout — so when Local Network permission is denied, which is exactly what the give-up message is about, the loop spun for the full 60s. That same path prints the multi-line permission instructions on every call, so the explanation would have been buried under thousands of copies of another message. The loop now paces against the attempt window: at most 12 queries and 3 devicectl probes in the budget, instead of unbounded.

**`main` could have been mirrored from an untested commit.** The sync workflow checked out `ref: dev`, which resolves at job time — so two merges landing close together would let the gate approve commit A's Test run while the push carried commit B. It uses `workflow_run.head_sha` now. That makes out-of-order completion normal (the older run finds `main` already ahead), so that case is detected and reported as "nothing to do" rather than tripping the divergence check.

Also fixed: the devicectl fallback had drifted *outside* its exception guard, so a missing `xcrun` or an unwritable temp dir would abort a run whose app was already live on the TV; giving up returned `LaunchResult.succeeded()` with no URI, which the runner turns into "connection to device ended too early" — an unrelated error, the same confusion this path exists to remove, so it exits on the explanation instead; `globals.xcodeProjectInterpreter!` no longer force-unwraps into a bare null-check mid-launch; the mDNS timeout override warns instead of silently defaulting when unusable; `_resolveDeviceIp`'s doc comment had slid onto the method inserted above it; `tvos_cache.dart` still documented the retired `v1.0.0-flutter3.44.0` tag shape this release replaces; `TMPDIR` in CI has a fallback rather than becoming empty if unset; and six analyzer excludes for directories that do not exist here are dropped.

The language-version regression test compared its output against another call to `currentLanguageVersion()`, so it passed for any value that function returned. It asserts the marker's shape and then its components against the SDK instead. Its `isNot(contains('3.9'))` assertion is gone — 3.9 is only wrong by accident of which Flutter is pinned, and is the *correct* answer on an SDK whose Dart is 3.9.

Four tests added for the timeout override parsing (default / valid / non-numeric / non-positive), which had none.

## Re-verified after the review round

`dart analyze lib/ test/` clean, **353 tests pass** (up from 349), and device-debug re-run on the **physical Apple TV 4K** — 13/13, 0 failed:

| | |
|---|---|
| launch · VM Service · heartbeat | ✅ |
| hot reload / hot restart (*effective*, not just reported) | ✅ |
| platform identity | ✅ `os=tvos isIOS/isTvOS/ffi all true, target=iOS` |
| FFI symbols | ✅ 10 exported from `Runner.debug.dylib` |

On real hardware: RWX JIT path active, Impeller/Metal up, `ffiMachine=AppleTV14,1 ffiSimulator=false`, hot reload landing 1 of 772 libraries. The CLI snapshot stamp matched the commit, so this was not a stale-snapshot pass.

**Scope of that run, stated plainly:** the VM Service resolved on the first attempt, so it proves the restructure did not regress the happy path — it does *not* exercise the busy-loop fix. That branch only occurs when macOS denies Local Network permission, which requires revoking it by hand. The pacing, the `throwToolExit` give-up and the devicectl time-throttle remain unverified end-to-end.

## Not addressed

- **The lldb stop-hook.** Upstream sets `target stop-hook add -o "thread backtrace all" -o "detach"`. tvOS debug/JIT needs a *persistent* debugger, so on the first stop lldb detaches and the session dies rather than becoming usable. It doesn't affect the happy path — hot reload works — but a crash ends the session instead of showing a Dart stack.
- **`_resolveReachableVmServiceUri` has no unit test.** It drives the `MDnsVmServiceDiscovery` singleton; covering it would mean injecting that global. Verified by repeated device runs instead.


jcourtois86 (approved):

MAUstaoglu:
## Added after approval — please re-review this part

`400efab` was folded in from #63's sibling PR (now closed) to save a review round. It is **unrelated to the 3.47.0 upgrade** and worth looking at on its own, because it changes repository policy as well as adding a file.

Flagging it explicitly because `dev` has `dismiss_stale_reviews: false`, so this push did **not** invalidate the existing approval — it would otherwise ride along unseen.

### What it does

Adds `.github/workflows/sync-main.yml`: after the **Test** workflow passes on `dev`, `main` is fast-forwarded to `dev`. No more `dev` → `main` pull request.

### Why the old dev → main PR kept conflicting

It wasn't bookkeeping — it was generating the conflicts. `main` required a linear history, and `dev`'s history is **not** linear:

```
b322013  Merge pull request #61 from fluttertv/chore/sync-main-into-dev
b7ca67a  Merge main into dev (1.5.1 release files)
8635711  Merge pull request #60 from fluttertv/chore/flutter-3.44.9
0e32f18  Merge main into dev
```

So that PR could only ever be **squashed** — a merge commit was rejected outright. A squash writes a commit onto `main` that exists in no part of `dev`'s history, `main` went one ahead, the branches genuinely diverged, and the next round conflicted on the version and `CHANGELOG` files. Two of the four merge commits above are the cleanup from previous rounds. A fast-forward can't reach that state: `main` becomes *literally* `dev`'s commit.

### Settings changed on `main` (outside this diff)

| setting | before | after | why |
|---|---|---|---|
| push allowlist | one user | *(none)* | the Actions token has to be able to write |
| `required_linear_history` | on | off | a mirror can't demand a stricter history than its source |
| force pushes | blocked | blocked | unchanged |
| deletions | blocked | blocked | unchanged |
| enforce on admins | on | on | unchanged |

The allowlist had to be dropped rather than extended: legacy branch protection accepts only installed GitHub Apps there and silently ignored the Actions bot (the POST returned success and added nothing), and a repository-level ruleset refuses it outright — `Actor GitHub Actions integration must be part of the ruleset source or owner organization`. An org-level ruleset would have kept `main` closed to humans, but that needs `admin:org` on the CLI token.

**Net effect worth naming:** `DenisovAV` and `jcourtois86` can now push to `main` directly, where previously only `MAUstaoglu` could. Force-push and deletion are still blocked.

### Safety in the workflow

- Triggers only on a **successful** Test run on `dev`, so a red `dev` never reaches the release branch.
- The push is **not** forced — git is what guarantees the fast-forward, rather than the script's own check being the only thing between a mistake and a rewritten `main`.
- It checks `main` is an ancestor of `dev` anyway, so a direct push to `main` fails the job with a message naming that cause instead of a bare rejection.
- `concurrency: sync-main` so two quick merges to `dev` can't race.

### Current state

`main` was fast-forwarded to `dev` by hand already — the two are `identical` at `b322013`. The workflow only fires on future pushes to `dev`, so that catch-up wasn't something it would have done.

MAUstaoglu:
Ready to land. Review round + device verification are folded into the description; summarising the state here.

**Head `fdc2ba4`** — `MERGEABLE`, `CLEAN`, `APPROVED`, `test` ✅ `analyze` ✅.

- `dart analyze lib/ test/` clean, **353 tests** pass.
- **Physical Apple TV 4K, debug: 13/13.** Launch, VM Service, heartbeat, hot reload and hot restart (verified *effective*), platform identity, and 10 FFI symbols exported. Snapshot stamp matched the commit, so not a stale-snapshot pass.
- 12 review findings fixed in `fdc2ba4` — details in the description.

One limitation worth having on the record: the device run resolved the VM Service on the first mDNS attempt, so it confirms the loop restructure did not regress the happy path but does **not** exercise the busy-loop fix. That branch only fires when macOS denies Local Network permission.

## Merge order

This must merge with a **merge commit** — `main` mirrors `dev` by fast-forward, and a squash would write a commit onto `main` that exists nowhere in `dev`, which is the divergence the mirror exists to eliminate. Squash and rebase are disabled on `fluttertv/engine` for the same reason; this repo still allows squash, so it has to be chosen deliberately here.

Merging this is also what **activates the mirror** — `sync-main.yml` lands on `dev` in the same push that triggers `Test`, so the workflow's first run is the one that syncs this release to `main`. `main` is currently `identical` to `dev`, so that first run is a clean fast-forward of exactly this PR's commits.

## After it lands

`main` picks it up automatically. The release is then cut by pushing a tag, which `release.yml` picks up (`v*.*.*`):

```
v3.47.0-tvos.1.6.0
```

following the established `v<flutter>-tvos.<cli>` shape (`v3.44.9-tvos.1.5.1`, `v3.44.8-tvos.1.5.0`, …), **not** a bare `v1.6.0`.

Two things to know about that workflow before the tag goes up:

1. It re-runs the full suite and `analyze` before creating the release, so a red tag build fails loudly rather than shipping.
2. It titles the release `flutter-tvos <tag>` — i.e. `flutter-tvos v3.47.0-tvos.1.6.0`. Every existing release reads `flutter-tvos 1.5.1 (Flutter 3.44.9)` instead, so those titles were set by hand. Worth a follow-up to make the workflow generate the established form rather than fixing each release after the fact.

MAUstaoglu:
Two updates since the review round, both from the "what's the minimum tvOS version now?" question — which turned out to catch a real bug rather than a docs nit.

## The deployment floor was wrong (`8dd2949`)

Upstream raised `ios_deployment_target` from `13.0` to `15.0` between Flutter 3.44.8 and 3.47.0, and the tvOS engine takes its deployment target from that same variable. The shipped `Flutter.framework` declares `minos 15.0` in all four variants; the previous engine declared `12.0`. **Nothing on our side moved with it.**

So `flutter-tvos create` produced an app declaring tvOS 13 that embeds a framework refusing to load below 15. It builds and runs on any current Apple TV — which is exactly why the hardware run passed and nothing flagged it. The failures are a user on an un-updated OS, and App Store validation rejecting the archive for the `MinimumOSVersion` / `minos` mismatch (ITMS-90208 — the very thing `tvosVersionMinFlag` exists to prevent).

Two of the ten declarations were **production constants feeding the compiler**, not documentation:

- `_kTvosMinimumOSVersion` in `application.dart`, which stamps `-mtvos-version-min` onto the AOT `App.framework`
- `TvosSwiftPackageManager.kDefaultDeploymentTarget`, i.e. `.tvOS("15.0")` in the generated SwiftPM manifest

plus the app template (Podfile + all three `TVOS_DEPLOYMENT_TARGET` configurations), the plugin podspec/`Package.swift` templates, the bundled package and its example, the docs, and 8 test assertions — 34 replacements across 19 files.

**No Apple TV hardware is dropped:** every model that can run tvOS 13 can run 15 or later, so this reaches only devices that have not been updated. Existing projects keep whatever their `tvos/` directory declares; `flutter-tvos create .` picks up the new floor.

Deliberately untouched: `PROGRESS.md` (a record of what was done, not a claim about now), and the porter's `@available` mirroring (it copies the *source plugin's* iOS versions, unrelated to our floor). Left for the next engine build: the simulator metallib is still compiled `-mappletvsimulator-version-min=13.0`, whose comment claims it matches the app template — now stale, but harmless, since a lower metallib floor is permissive rather than restrictive.

## Analyzer excludes (`54c40e0`)

An earlier commit on this branch dropped the per-platform excludes for matching nothing. Right observation, wrong conclusion — they are back and now anchored with `**/`.

Analyzer excludes are globs relative to `analysis_options.yaml`, so the bare forms only ever matched a top-level directory. `build/**` therefore missed `packages/flutter_tvos/build/`, the one generated directory this repo actually produces. `**/build/**` covers it, and the platform entries become forward-looking rather than decorative — an example app regains `ios/`, `android/` and friends the moment anyone runs `flutter create` in one, and a generated host project is not ours to lint.

Verified `dart analyze lib/` (what CI runs) still exits 0, and a whole-package analyze no longer reports anything under `packages/flutter_tvos/build`.

## State

353 tests pass, analyze clean. Staying at **1.6.0** rather than 2.0.0: nothing breaks for existing projects, which keep their current `tvos/` settings.

The landing page carries the same floor change — fluttertv/landing#8 — and should merge only after this lands and the release is tagged, since its docs links point at `blob/main`.

DenisovAV (commented):
## Review

Reviewed the whole diff against `flutter_tools` at the pinned `4cf24164`. The 3.47.0 re-mirroring is right, the mDNS retry design is right, and `sync-main.yml` is careful in exactly the places that matter. **One required change** before merge, in one function — everything else I looked at came back clean, including three things I initially suspected and then disproved.

---

### Required: the devicectl probe short-circuits the retry loop with a host that cannot resolve

`lib/tvos_device.dart:1223-1230`. `probedDevicectl` starts `false`, so `_resolveDeviceIp` runs on the **first** iteration, right after the first 5 s mDNS attempt — and if it returns anything, `_resolveReachableVmServiceUri` returns immediately.

The PR body argues this is safe:

> a LAN-paired Apple TV reports `tunnelState: "disconnected"` with no `localHostnames` and no `ipAddress`, so `_resolveDeviceIp` returns null every time

Those two fields are indeed empty — but `_resolveDeviceIp` reads a third one **between** them, `connectionProperties.potentialHostnames` (`:1306`), and that one is populated. From the paired Apple TV on my Mac right now:

```
ВУФ | tvOS
  tunnelState       : unavailable
  networkAddresses  : None
  localHostnames    : None
  potentialHostnames: ['VUF.coredevice.local',
                       'A0416F11-8EB2-5767-82FB-2EF1E2E6C606.coredevice.local',
                       '2b0b4c6a0efcbffab378bc07e0fdf30524bb570b.coredevice.local']
```

So `_resolveDeviceIp` returns `VUF.coredevice.local`. And that name does not resolve:

```
VUF.coredevice.local                                    FAILS — gaierror
A0416F11-8EB2-5767-82FB-2EF1E2E6C606.coredevice.local   FAILS — gaierror
Sashas-MacBook-Pro.local                                ::1      ← control, resolves
remoted: running    remotepairingd: running
```

The control `.local` name resolving proves the resolver itself is fine, and both CoreDevice daemons are up. `*.coredevice.local` records are registered locally by `remoted` and only exist **while a tunnel to that device is up** — with `tunnelState` anything other than connected, they are dead names for every process on the machine, independent of Local Network permission.

Net effect: `http://VUF.coredevice.local:<port>/<token>/` is handed back as "reachable", the caller takes any non-null URI as success, the new 60 s budget and the give-up `throwToolExit` are never reached, and the resident runner loops on a failed host lookup. That is the silent-dead-hot-reload this PR exists to remove.

It is not limited to the denied-permission case either. **Any** first-attempt mDNS miss reaches the probe — including the ordinary Bonjour propagation race that is this PR's primary target. Since `potentialHostnames` is populated for every paired device, the retry machinery is effectively unreachable on the failure path.

Worth saying why 13/13 green runs didn't catch it: on a Mac with Local Network granted and a warm cache, mDNS wins on attempt 1 and returns before the probe, so this branch never executed. The runs are real evidence for the happy path and say nothing about this one.

**Suggested fix** — keep the fast path for a numeric address, verify a hostname before trusting it:

```dart
if (!probedDevicectl ||
    elapsed.elapsed - lastDevicectlProbe >= _devicectlProbeInterval) {
  probedDevicectl = true;
  lastDevicectlProbe = elapsed.elapsed;
  final String? deviceHost = await _resolveDeviceIp(deviceId);
  // devicectl yields either a numeric address from networkAddresses or a
  // *.coredevice.local name from potentialHostnames. The latter is registered
  // locally by remoted and only exists while a tunnel to the device is up, so
  // on a LAN-paired Apple TV it is a dead name — and returning it here would
  // bypass the whole retry budget below with a URI nothing can dial.
  // A numeric address costs nothing to confirm; a name that does not resolve
  // is discarded so the loop can keep waiting for Bonjour.
  if (deviceHost != null && await _hostResolves(deviceHost)) {
    logger.printTrace('Resolved device host via devicectl: $deviceHost');
    return deviceUri.replace(host: deviceHost);
  }
}
```

```dart
Future<bool> _hostResolves(String host) async {
  try {
    final List<InternetAddress> addresses = await InternetAddress.lookup(
      host,
    ).timeout(const Duration(seconds: 2));
    return addresses.isNotEmpty;
  } on Object {
    return false;
  }
}
```

Dropping the `potentialHostnames` branch from `_resolveDeviceIp` outright would also work and would match what the PR body already argues ("devicectl cannot rescue it") — the code just doesn't currently match that argument. Either is fine by me; the lookup keeps the branch useful for the tunnel-connected case.

---

### Also worth fixing, one token

`test/README.md:31` and `:34` document `TMPDIR="$(cd "$TMPDIR" && pwd -P)"` with no default, while both workflows correctly use `${TMPDIR:-/tmp}`. With `TMPDIR` unset — a plain `ssh` session to a Mac, for instance — `cd ""` succeeds without changing directory, so `pwd -P` prints the repo root and the command exports `TMPDIR=<repo root>`. The FS guard then treats the whole working copy as the allowed temp root, which is the opposite of what the paragraph below promises. `${TMPDIR:-/tmp}` in both lines.

---

### Checked and clear

Things I went after and could not make stick:

- **Simulator regression.** The new block also triggers on `127.0.0.1`, which is a *valid* host for a simulator, so a hard `throwToolExit` there would have killed simulator debugging. It doesn't: `startApp` branches to `_startAppOnSimulator` at `:564` and returns long before this code. No overlap.
- **`throwToolExit` on a null `XcodeProjectInterpreter` (`:750`).** I suspected it could abort with the `--start-stopped` app and the `devicectl --console` child still live, skipping the Xcode fallback and teardown. The branch is unreachable under the tool runner — `context_runner.dart:359` registers `XcodeProjectInterpreter` unconditionally and the generator constructs one whether or not Xcode is present. Defensive dead code, not a leak.
- **`throwOnMissingLocalNetworkPermissionsError: false`.** I expected the argument that retrying is pointless because permission can't be granted to a running process. Apple's TN3179 says the opposite — the system picks up a mid-run grant and Apple explicitly recommends retry logic — and each attempt here opens fresh sockets. The design is right. The only residual is cosmetic: upstream prints its 8-line instruction block on every call (`mdns_discovery.dart:275`), so a denied run shows ~12 copies before the give-up message. Worth deduping someday, not now.
- **Pins.** `engine-495915c…` is published with all six variants, and that commit survived the engine merge — it is an ancestor of `fluttertv/engine@main`, so the tag still names a reachable revision. `flutter.version` `4cf24164` is the `3.47.0` tag.
- **`recorded_uses.json`.** Matches upstream `KernelSnapshot.build` including the `buildMode.isPrecompiled` split; `extraFrontEndOptions` is growable on both construction paths, so the `.add` is safe.
- **`LinkHooks.loadHookResult`** is the same call upstream's `CopyFlutterBundle.build` makes.
- **`throwOnMissingLocalNetworkPermissionsError` exists** in 3.47.0 (`mdns_discovery.dart:172`, default `true`), so the override is real API, not a guess.
- **`sync-main.yml`.** Checks out the tested SHA rather than `dev`, handles "main already contains this" for out-of-order runs, refuses to move main backwards, and pushes without `--force` — so the fast-forward property is enforced by git itself rather than only by the checks above it. This is the part I'd have most expected to get wrong and it doesn't.
- **README** no longer overclaims: the on-device debug paragraph now leads with "needs a debugger attached", which matches how tvOS actually gates JIT.

---

### One thing to flag on handover, not to change

An mDNS miss used to degrade (session alive, hot reload quietly dead); it is now a hard `throwToolExit`. For a human that is strictly better. For anything automated running `flutter-tvos run` against a device — a CI box without Local Network permission, say — it turns a partially-working run into a failed one. Right call, but worth a sentence to whoever operates that, since the CHANGELOG files it under *Fixed*.

Happy to approve as soon as the resolve check lands.

lib/tvos_device.dart
```diff
@@ -1064,9 +1147,115 @@ class TvosDevice extends Device {
...
+        // and losing the VM Service host must not abort a run whose app is
+        // already installed and running on the TV.
+        if (!probedDevicectl ||
+            elapsed.elapsed - lastDevicectlProbe >= _devicectlProbeInterval) {
+          probedDevicectl = true;
+          lastDevicectlProbe = elapsed.elapsed;
+          final String? deviceHost = await _resolveDeviceIp(deviceId);
+          if (deviceHost != null) {
+            logger.printTrace('Resolved device host via devicectl: $deviceHost');
+            return deviceUri.replace(host: deviceHost);
```
DenisovAV, line 1251: This is the required change — see the main comment for the evidence.

`probedDevicectl` starts `false`, so this runs on iteration 1 and an early `return` here bypasses the entire 60 s budget, the pacing, and the give-up `throwToolExit`.

`_resolveDeviceIp` can return a `*.coredevice.local` name from `potentialHostnames` (`:1306`) — populated for every paired device, including the Apple TV on my Mac, where `localHostnames` and `networkAddresses` are both empty as the PR body says. That name only resolves while a CoreDevice tunnel is up; with `tunnelState: unavailable` it fails `getaddrinfo` for every process on the machine. Returning it hands the resident runner a URI it can never dial.

Guard the return on the host actually resolving — a numeric address from `networkAddresses` passes instantly, a dead name falls back into the retry loop where it belongs.

test/README.md
```diff
@@ -28,12 +28,33 @@ test/
 
 ```bash
 # Run all tests
-flutter/bin/dart test test/
+TMPDIR="$(cd "$TMPDIR" && pwd -P)" flutter/bin/dart test test/
 
 # Run a specific test file
-flutter/bin/dart test test/general/tvos_emulator_test.dart
+TMPDIR="$(cd "$TMPDIR" && pwd -P)" flutter/bin/dart test test/general/tvos_emulator_test.dart
```
DenisovAV: `${TMPDIR:-/tmp}` here, to match `test.yml` and `release.yml`.

With `TMPDIR` unset — a plain `ssh` session to a Mac — `cd ""` succeeds without changing directory, so `pwd -P` prints the repo root and this exports `TMPDIR=<repo root>`. The FS guard then accepts writes anywhere in the working copy, which is the opposite of what the paragraph below promises.

DenisovAV (approved):

PR #69, opened 2026-08-21
Retarget the CLI to Flutter 3.32.8

DenisovAV:
Pinning `flutter.version` to 3.32.8 left the tool **uncompilable** — 15 errors. The CLI is written against a newer `flutter_tools` than 3.32.8 ships. This resolves all of them in the direction of what 3.32.8 actually provides.

## The five incompatibilities

| | what 3.32.8 lacks | resolution |
|---|---|---|
| `application.dart` | `DartHooksResult`, `DartBuild.loadHookResult`, `copyAssets(dartHookResult:)` | dropped — we already skip native-assets for tvOS and write an empty manifest, so there were never hook results to pass |
| `tvos_device.dart` | `src/ios/lldb.dart` entirely | removed the lldb attach; fall through to the Xcode debugger this file already implements |
| `tvos_device.dart` | `Device.isSupported` is sync here | `bool` instead of `Future<bool>` |
| `executable.dart` | `UpdatePackagesCommand(verboseHelp:)` | no constructor args here |
| `tvos_plugins.dart` | — | registrant hardcoded `// @dart = 3.9`; Dart 3.8.1 rejects it |

The last one only surfaced at **build** time, not compile time, and is the one worth keeping an eye on: it is now derived via `currentLanguageVersion(...)` rather than hardcoded, so it tracks whatever Flutter is pinned instead of breaking on the next move in either direction.

Two tests asserted newer-SDK specifics: the `informative` development artifact (absent here) and `writePackageConfigFiles` (`writePackageConfigFile` here).

## Device debug is not available on this line

Worth stating plainly, because it is blocked twice over and neither is fixed here:

1. `flutter_tools` gained the LLDB helper after 3.32.8, so Path 1 of the device debug flow cannot be built at all.
2. The published `v1.0.1-flutter3.32.8` engine carries both the RWX marker and Dart's `Must run with --write-protect-code` assertion — verified by inspecting the shipped binary. `CheckIfNeedDebuggerHelpWithRX()` `FATAL`s when the flag is off on a new enough OS, which is exactly the state our own `01-dart-globals.patch` creates.

On (2) I could not verify one link: `IsAtLeastIOS18_4()`'s definition is not in the pinned Dart tree where it is used, so whether it returns true on tvOS 26 rests on the abort reported in the engine README rather than on code I read. Also note `flutter3.32.8/patches/` contains **no** `!defined(DART_HOST_OS_TVOS)` exclusion, while `flutter3.44.8/patches/` does — so the claim in the engine README that the archived set carries this fix looks wrong and is worth its own issue.

The Xcode debugger fallback remains, and the simulator is unaffected (JIT works there without a debugger).

## Verified

- `flutter-tvos --version` → Flutter 3.32.8 • Dart 3.8.1
- `create --platforms=tvos` → project generated
- `build tvos --simulator --debug` → Xcode build done, 17.6s
- `build tvos --simulator --release` → Xcode build done, 17.5s (AOT + icon tree-shake)
- `dart analyze lib/` → 0 errors, 0 warnings (250 pre-existing `always_specify_types` infos, in untouched files)
- `flutter test` → **346/346 pass**

## Follow-up, not fixed here

`tool_revision()` in `bin/internal/shared.sh` returns the git HEAD SHA in a git checkout, so **uncommitted edits under `lib/` never invalidate the snapshot** — the file-hashing fallback only runs for non-git checkouts. Editing the tool and re-running it silently executes the old snapshot.


DenisovAV:
Closing: targeting `dev` is wrong. `dev` is on Flutter 3.47.1; this branch pins 3.32.8, so merging it would move the whole toolchain backwards. The repo has no parallel-version model — the `chore/flutter-*` branches all step forward into `dev`.

Splitting instead: the one version-agnostic fix here (deriving the plugin registrant's language version instead of hardcoding `// @dart = 3.9`) goes to `dev` on its own, and the 3.32.8 retarget is rebuilt on top of current `dev` so it carries the archive-mode guard and the rest of #66 rather than branching from a three-week-old base.

PR #73, opened 2026-08-29
Sign the engine with the developer's own Developer ID; embed it explicitly; Flutter 3.47.2

MAUstaoglu:
## Engine signing (ITMS-91065)

Apple treats Flutter as a commonly-used third-party SDK and rejects submissions whose embedded engine is not signed by the SDK's provider. **The engine must carry a `Developer ID Application` signature before the build embeds it.** Signing the copy inside the app does not satisfy the check, and neither does the Distribution signature `xcodebuild -exportArchive` applies.

That last point is counter-intuitive, so it was measured: `-exportArchive` re-signs nested frameworks unconditionally, and a **stock Flutter iOS app** ships its engine signed `Apple Development: <the developer>` with no trace of `FLUTTER.IO LLC` — and passes. So the authority on the embedded copy is not what Apple reads.

What decided it, same app and pipeline, one afternoon:

| Builds | Engine artifact | Outcome |
|---|---|---|
| 4 builds across Flutter 3.41.4, 3.44.5, 3.47.2 | unsigned | ITMS-91065 within ~2 min of external submission |
| 2 builds, 3.47.2 | Developer ID signed | cleared the automated check |

The bundle identifier made no difference (one rejected build shipped `dev.fluttertv.Flutter` and was rejected naming the same path).

### What this PR does

Published artifacts stay **unsigned** — no maintainer identity in a public release, no release gated on one person's certificate. `TvosEngineSigner` signs locally instead, from `NativeTvosBundle.build()` at step 0b, before `_copyFlutterFramework` (the copy preserves whatever signature the cache holds at that moment). It signs the extracted cache, covering both the explicit embed phase and a legacy project's SwiftPM binary target that symlinks straight into it.

- Signs **by SHA-1 hash, never display name** — duplicate certificates sharing a name are common, and `codesign --sign <name>` then fails with `ambiguous (matches … and …)`.
- `--timestamp --options runtime`, matching how flutter.dev signs its own engine.
- Device builds only; re-signs only when the identity differs.
- **Never fails the build** — a missing Developer ID warns and names ITMS-91065, since only submission needs it.
- `TVOS_ENGINE_SIGNING_IDENTITY` / `TVOS_ENGINE_SKIP_SIGNING` to override or disable.

⚠️ **For users:** Apple lets only a team's **Account Holder** create a Developer ID certificate, with a per-team limit. An Admin on someone else's team will need one made for them. Documented in the new README "Code signing" section.

**Untested:** whether an `Apple Distribution` certificate would also satisfy the check. The signer deliberately does not fall back to one — a wrong-type signature is not obviously better than none and would hide the real problem — but if it works, that removes the Account Holder requirement.

## Explicit embed pipeline

Also in this branch: the engine is embedded by the app itself rather than by Xcode from a SwiftPM `.binaryTarget`, matching what upstream Flutter returned to in [flutter/flutter#181739](https://github.com/flutter/flutter/pull/181739). The generated `FlutterFramework` package is empty; a "Prepare Flutter framework" scheme pre-action stages the engine into `BUILT_PRODUCTS_DIR`, and an "Embed Flutter.framework" phase puts it in the bundle. Pre-existing projects keep the old path and are warned.

## Flutter 3.47.2

Pins `d3b14c876900e553bc736ca19295fc09e3853e8e` and engine [`engine-27c707e8b2d18b3b50cd555cd88ab9b2a01f53dc`](https://github.com/fluttertv/engine-artifacts/releases/tag/engine-27c707e8b2d18b3b50cd555cd88ab9b2a01f53dc). Requires the companion PR fluttertv/engine#19 — those artifacts use `dev.fluttertv.Flutter` and are not interchangeable with an older CLI.

## Testing

- 11 new unit tests for the signer; full suite **426 passing**, `dart analyze` clean.
- Simulator gate green: launch, hot reload/restart effective, platform identity, 10 FFI symbols.
- Device matrix green on a physical Apple TV (debug/profile/release).
- Verified end-to-end by stripping every signature from the cache and confirming an ordinary `flutter-tvos build tvos --release` re-applies it, through to the exported IPA.

**Not yet confirmed:** the submissions cleared Apple's *automated* check but none has reached `BETA_APPROVED`. Worth waiting for that before tagging a release.


DenisovAV:
Reviewed with fluttertv/engine#19, where the engine-side notes are.

## Revoked certificates are selected silently

`security find-identity -v -p codesigning` annotates revoked certs `(CSSMERR_TP_CERT_REVOKED)` and still counts them in "N valid identities found" — two of ten on this machine. The regex at `tvos_engine_signing.dart:62` does not reject the suffix, and `:131-139` filters by name prefix before taking `.first`.

Driven against the real parser with a synthetic revoked `Developer ID Application` line: selected when listed first, selected when it is the only one, accepted when named via `TVOS_ENGINE_SIGNING_IDENTITY`. Real `codesign` with such a cert and the signer's exact flags exits 0. Only trace is a `printTrace`.

The CLI's only `codesign --verify` runs on the embedded copy *after* the app re-sign, so it never sees the revoked signature. And because a revoked cert shows `Authority=(unavailable)`, `_isSignedBy` returns false and the cache is re-signed on every build.

The comment at `:137` — "Duplicates here are the renewed-certificate case … either will verify" — does not hold for a revoked cert. The same function never consults `DEVELOPMENT_TEAM`, so with two Developer IDs installed, project B can get team A's certificate.

## Deleting the signing call entirely keeps the suite green

13 mutations inside `TvosEngineSigner` were all killed — the unit layer is solid. One level up, each of these leaves 426/426 passing:

| mutation | result |
|---|---|
| delete the `_signEngineForDeveloper()` call from `build()` | 426 pass |
| move it after `_copyFlutterFramework`, breaking the stated ordering | 426 pass |
| remove the `!buildInfo.simulator` guard | 426 pass |
| strip the sign block from "Embed Flutter.framework" | 426 pass |

No test constructs `NativeTvosBundle` and calls `.build()`; `_signEngineForDeveloper` appears nowhere under `test/`.

The fourth survives because `tvos_app_bundle_test.dart:406-407` asserts `contains('codesign --force --sign')` over the whole pbxproj, and the unrelated "Embed App.framework" phase supplies that string by itself — confirmed by stripping that phase too, at which point the assertions fail. `stripJitPayload` and `pbxprojLacksBuildModeGuard` are the pattern to follow: static helpers tested directly.

## Merging into `main` breaks `Sync main`

Simulated both merge styles: after either, the next dev commit hits `::error::main … is not an ancestor` and exits 1. "Tag release" chains off a successful Sync main, so the 1.9.0 tag goes with it. Fast-forwarding dev to main's new tip before the next dev push avoids it.

## Two incompatible accounts of what ITMS-91065 reads

**M1** — the Developer ID must be on the engine before embedding; an app-identity re-sign of the embedded copy does not satisfy Apple. `CHANGELOG.md:10-13`, `engine/build.sh:814-819`.

**M2** — Apple reads the app-identity signature on the embedded copy and that suffices. `CHANGELOG.md:69-72`, `test/general/tvos_app_bundle_test.dart:369-373`, `project.pbxproj.tmpl:238`, plus six places in the engine repo.

`27c707e`'s commit message says it *"corrects the neighbouring comment, which claimed an app-identity re-sign of the embedded copy was what satisfied ITMS-91065"* — M2 is already settled as wrong. One occurrence was fixed; the rest stand, including the template comment that ships into every generated project, and `CHANGELOG.md`, where M1 and M2 sit sixty lines apart from two commits of this PR.

`README.md:162` says "signed by the SDK's provider", which matches neither account nor `tvos_engine_signing.dart:1`.

## The staged engine and the shipped engine come from different paths

The pre-action stages `PROJECT_DIR/Flutter/Flutter.framework` into `BUILT_PRODUCTS_DIR` — what plugin targets compile against — while the Embed phase copies from `PROJECT_DIR`. Upstream reads its embed from `BUILT_PRODUCTS_DIR` and stages from a Run Script phase as well as a pre-action, so a missed pre-action self-heals on the next build; here it does not.

Narrow in practice: needs an Xcode-started build, the pre-action entirely absent (custom scheme, or `xcodebuild -target`), and stale DerivedData. Nothing detects it when it happens. The scheme test checks the pre-action exists but not that its `BlueprintIdentifier` matches the pbxproj.

## Measurement worth having on record

Ran the template's embed script against the real artifact in two states, Developer-ID-signed and unsigned, and diffed the output. `Info.plist`, `CodeResources`, the Mach-O before `LC_CODE_SIGNATURE`, the SuperBlob and the CodeDirectory are identical. The only difference in the shipped bundle is 4,156 bytes of unreferenced padding past the blob's declared length, naming no signer.

No test covers the property, so a future change to the embed phase can remove whatever preserves it without anything failing until a rejection days later.

## Verified good

3.47.2 sync: zero patch files changed; upstream's 17 commits across 44 files touch none of the 58 the patch set does. Details on #19.

Rename: `strings` on the shipped binary finds no bare `io.flutter.flutter`.

MAUstaoglu:
@DenisovAV Thanks — this was a genuinely useful review, and the mutation testing in particular found things our own tests could not. All four findings are addressed in `718e2db`, `2452a03`, `6e16955`.

## Revoked certificates

Fixed. The parser used `firstMatch` with nothing anchoring end-of-line, so the annotation was simply ignored. It now anchors and rejects *any* trailing annotation, not only `CSSMERR_TP_CERT_REVOKED` — an annotation we don't recognise is a reason to skip a certificate rather than trust it.

Three tests cover the cases you drove (revoked first, revoked as the only Developer ID, revoked named explicitly by hash). We reverted the parser to confirm all three fail against the old code, so they pin the behaviour rather than passing incidentally. The `:137` duplicate-certificate comment holds again now that revoked entries are gone before that logic is reached.

`DEVELOPMENT_TEAM` is deliberately **not** addressed here. You're right that with two Developer IDs installed project B can get team A's certificate, but that's a design decision rather than a defect, and we'd rather make it separately than fold it into this PR.

## The call site

Fixed, and your diagnosis of why the fourth mutation survived was exactly right.

`build()` runs `pod install` and `xcodebuild`, so driving it whole would have been brittle. Instead we extracted `signAndStageEngine` as a `@visibleForTesting` seam — signing and staging together, because the order *between* them is the invariant. No marker files were needed: `_copyFlutterFramework` shells out to `cp -R`, and `FakeProcessManager.list` enforces command order, so `security` → `codesign` → `cp` is directly observable.

All four mutations now fail the suite:

| mutation | before | now |
|---|---|---|
| delete `_signEngineForDeveloper()` from `build()` | 426 pass | fails |
| move it after `_copyFlutterFramework` | 426 pass | fails |
| remove the `!buildInfo.simulator` guard | 426 pass | fails |
| strip the sign block from "Embed Flutter.framework" | 426 pass | fails |

Two things worth reporting back:

**Our first simulator test passed a mutated build.** We'd seeded only the simulator cache, so dropping the guard hit `signIfPossible`'s early return on a missing directory instead of attempting to sign — green for the wrong reason. Both variants are now seeded, and the helper carries a comment explaining why.

**Scoping the pbxproj assertions exposed a live error in the existing test.** It asserted `$(PROJECT_DIR)/Flutter/Flutter.framework` with parens; the phase's script uses `${PROJECT_DIR}` with braces. The paren form was being satisfied by the phase's own `inputPaths`. Both are real, so they're now asserted separately — `inputPaths` for Xcode's dependency tracking, `${...}` for what the shell reads.

The `BlueprintIdentifier` assertion is in, verified by renaming the Runner target id and watching it fail naming both files.

## The two accounts of ITMS-91065

Swept across both repos — no M2 phrasing left in either. The `project.pbxproj.tmpl` comment that ships into every generated project was the worst of them.

The correction kept M2's *conclusion* where it was right and replaced its *reason*: origin signing is optional because the CLI signs locally before Xcode embeds, not because an app-identity re-sign satisfies the check.

We deliberately did **not** rewrite those comments to say the embed-phase signature is pointless — it isn't, since a framework reaching the bundle unsigned fails to load and `-exportArchive` re-signs regardless. They now state it's the ordinary app-identity signature, *separately* from the fact that it isn't what satisfies ITMS-91065.

## `Sync main`

Good catch, and it would have bitten. This PR now targets `dev` rather than `main`, so the mirror stays intact and `Tag release` can fire. We also confirmed the flutter pin resolves to `3.47.2`, so the tag will be `v3.47.2-tvos.1.9.0` and the version-resolution step won't hard-fail.

## Your note, not a finding

Taken seriously, and left as-is deliberately. Since nobody can currently name what preserves the property, we avoided asserting a mechanism anywhere in the new comments — they state the observed A/B and nothing more. You're right that a regression there would surface as a rejection days later with no test catching it.

One thing to flag honestly: we're proceeding to tag without waiting for `BETA_APPROVED`, against your suggestion. That's a maintainer call rather than a disagreement with your reasoning.

## Residual gap

A mutation deleting `signAndStageEngine(...)` from `build()` itself would still pass. The seam shrinks the untested surface to a single call rather than eliminating it; covering that honestly needs a full `build()` drive, which didn't look worth it here. Flagging it rather than leaving you to find it.

DenisovAV (approved):
Verified against `718e2db`.

The parser now anchors and rejects any trailing annotation rather than only `CSSMERR_TP_CERT_REVOKED`, which is the right generalisation — an annotation we don't recognise is a reason to skip a certificate, not to trust it. Reverting the `m.group(3) != null` rejection fails three tests, so they pin the behaviour rather than passing incidentally.

All four mutations that survived the previous revision are dead:

| mutation | result |
|---|---|
| delete `_signEngineForDeveloper()` from the seam | fails (−1) |
| move it after `_copyFlutterFramework` | fails (−1) |
| remove the `!buildInfo.simulator` guard | fails (−1) |
| strip the sign block from the Embed phase | fails (−4) |

The residual gap you flagged is real — deleting `signAndStageEngine(tvosProjectDir)` from `build()` leaves all 104 green. Reasonable to leave: a full `build()` drive costs more than the single call it would cover.

Retargeting to `dev` resolves the mirror breakage, and the pin resolves to 3.47.2, so `Tag release` produces `v3.47.2-tvos.1.9.0`.

Splitting the `PROJECT_DIR` assertions is correct — `$(...)` against `inputPaths` for Xcode's dependency tracking, `${...}` against the script text for what the shell reads.

One note for anyone running the suite locally: 36 tests in `tvos_app_bundle_test.dart` fail under `fs_safety.dart`'s `createTempSync` guard. Not from this PR — the same 36 fail on `dev`. With `FLUTTER_TEST_DISABLE_FS_GUARD=true` the four affected files are 104/104.

Deferring `DEVELOPMENT_TEAM` to its own change is reasonable; it is a design question rather than a defect in this one.

PR #75, opened 2026-09-03
Make the compiled snapshot see uncommitted CLI sources

MAUstaoglu:
## The problem

`tool_revision` in `bin/internal/shared.sh` decides whether `bin/cache/flutter-tvos.snapshot` is stale. Its git branch returned `git rev-parse HEAD` — which describes what was **committed**, not what is on disk.

In a development checkout those differ constantly. Edit anything under `lib/`, run the CLI, and it silently executes the *previously compiled* snapshot.

The failure mode is why this is worth fixing rather than documenting. The change does not appear to be ignored — it appears to be **wrong**. You get the same build, the same error, byte for byte, and conclude the fix is incorrect when in fact it never ran. `rm bin/cache/flutter-tvos.snapshot` is the folk remedy, for anyone who has been told it.

It cost me a full build cycle and a byte-identical error I nearly misread as "my patch doesn't work".

## The fix

The **non-git branch of this same function never had the problem** — it hashes the contents of `bin/` and `lib/`, so it is content-addressed by construction. This makes the git branch agree, by folding uncommitted work into the revision.

- **Scoped** to `bin`, `lib`, `pubspec.yaml`, `pubspec.lock`, so editing a README or a test does not force a recompile.
- **Hashes `status --porcelain` *and* `diff HEAD`.** Names and statuses alone would collide across two different edits to the same file; the diff alone would miss a newly added one.
- **Only paid when the tree is dirty.** A clean checkout returns the bare 40-character sha exactly as before, so nobody who is not editing the CLI sees any change at all.

## Tests

Nine cases in `test/general/tvos_snapshot_revision_test.dart`, driving the real bash function in a throwaway git repository: clean checkout returns bare HEAD; edits under `lib/` and `bin/` move it; a new untracked source moves it; two different edits to one file do not collide; the value is deterministic rather than a nonce; reverting restores it; README/test edits do **not** move it; committing returns to a bare sha.

**Six of the nine fail against the unfixed `shared.sh`.** The three that pass either way are the ones asserting unchanged behaviour, which is what they are for.

## Baseline

The suite is not currently green on `main`: **58 failures** there, and the same 58 here. Passes go 373 → 382, exactly the nine added.

## Note on versioning

No version bump here deliberately — #74 is open and bumps to 1.10.0, and a second bump would just conflict. This should ride whichever release lands next.

### DIFF
--- a/bin/internal/shared.sh
+++ b/bin/internal/shared.sh
@@ -30,7 +30,40 @@ DART_EXE="$FLUTTER_DIR/bin/cache/dart-sdk/bin/dart"
 
 function tool_revision() {
   if [[ -d "$ROOT_DIR/.git" ]] && git --git-dir="$ROOT_DIR/.git" rev-parse HEAD >/dev/null 2>&1; then
-    git --git-dir="$ROOT_DIR/.git" rev-parse HEAD
+    local head
+    head="$(git --git-dir="$ROOT_DIR/.git" rev-parse HEAD)"
+
+    # HEAD alone describes what was committed, not what is on disk. In a
+    # development checkout those differ constantly, and the difference is
+    # invisible here: edit anything under lib/, run the CLI, and it silently
+    # executes the previously compiled snapshot instead. The symptom is a
+    # change that appears to do nothing -- the same build, the same failure,
+    # byte for byte -- which reads as "my fix is wrong" rather than "my fix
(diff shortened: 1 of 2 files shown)


MAUstaoglu:
Folded into #74 so 1.10.0 lands as one release — the two fixes are the same piece of work, and the snapshot bug is what made the first one look broken while I was writing it. Same commits, same tests.

PR #76, opened 2026-09-08
Add --no-codesign to `build tvos`

MAUstaoglu:
Lets a device build compile on a machine with no signing certificate — which is every hosted CI machine.

Without it, `_resolveSigningArgs` falls through to an error and empty args, and `xcodebuild` fails with *Signing for "Runner" requires a development team*. Upstream `flutter build ios` has carried `--no-codesign` for the same reason.

`CODE_SIGNING_ALLOWED=NO` alone is not enough: Xcode still resolves a provisioning profile for the bundle's entitlements, which is the half that fails with no certificate. The flag also drops `-allowProvisioningUpdates` and the App Store Connect key — there is no profile to update and nothing to authenticate for.

### Verified, not assumed

On a real `appletvos` release build with no team in the environment:

| | |
|---|---|
| Mach-O | arm64, `LC_BUILD_VERSION platform 3` (tvOS device, not simulator) |
| signature | `code object is not signed at all`, no `embedded.mobileprovision` |
| FFI exports | all ten `_flutter_tvos_*` symbols defined **and** in the dyld export trie |

That last row is the one that matters: it is exactly the assertion the release train's compile-only gate makes, and it holds unsigned. A signed build with `DEVELOPMENT_TEAM` set still produces an embedded CodeDirectory, unchanged.

449 unit tests pass; `dart analyze --fatal-infos lib/ test/ bin/` is clean. Two tests added — one that the default stays `true` (a default of `false` would silently stop signing real device builds), one that provisioning args are dropped when unsigned.

### Ships as 1.10.1

`pubspec.yaml`, `README.md` and `CHANGELOG.md` are bumped, so merging this cuts `v3.47.2-tvos.1.10.1`.

Two documentation fixes ride along. `doc/commands.md` said a device build requires a development team, which is no longer true — it now says what each path needs and when to reach for the flag. And `doc/architecture.md`'s SPM note pointed at a section of a file that is not in the repository; it now names the "Prepare Flutter framework" pre-action and the "Embed Flutter framework" build phase in the Runner project.


DenisovAV:
Verified against `3bfdaf4`: 449 tests green.

The interaction with the embed phases is safe, and the ordering in the Flutter.framework phase is what makes it safe: the `[ ! -d "${FRAMEWORK_SRC}" ]` check runs first and unconditionally, so `--no-codesign` cannot suppress the one error that catches an app about to ship without an engine. Only the `codesign` call itself sits behind the signing guard. The App.framework phase requires `-n "${EXPANDED_CODE_SIGN_IDENTITY}"` for the same reason. Passing an empty identity therefore lands in the intended branch of each rather than invoking `codesign --sign ""`.

Suppressing the authentication args when nothing is being signed is right, and for the right reason — the key exists to let xcodebuild fetch a provisioning profile, and an unsigned build asks for none.

One finding.

## The five settings the feature consists of have no coverage

The comment above them states: "Every one of these is needed: `CODE_SIGNING_ALLOWED=NO` alone still leaves Xcode resolving a provisioning profile for the bundle's entitlements, which is the half of signing that fails on a machine with no certificate."

Nothing tests that. Mutating the source and re-running the full suite:

| mutation | result |
|---|---|
| drop `CODE_SIGNING_REQUIRED=NO` | **survives** |
| drop `CODE_SIGNING_ALLOWED=NO` | **survives** |
| drop `EXPANDED_CODE_SIGN_IDENTITY=` | **survives** |
| stop suppressing the auth args when unsigned | **survives** |
| ignore `--no-codesign` in `_resolveSigningArgs` entirely | **survives** |
| default `codesign` to `false` | fails |

The new test is a real assertion, but it exercises `xcodebuildArgs`, which receives `codesign` as a parameter — so it pins the plumbing (the flag reaches the argv builder and drops `-allowProvisioningUpdates`) and not the payload (which settings actually make an unsigned build work). The feature can be switched off inside the resolver and CI stays green.

That matters more than usual here because of where the failure would surface. These settings are only exercised on a machine with no certificate — which is not where anyone runs the suite, and is exactly the machine the flag exists for. A regression would land as a red Codemagic build in `release-train`, not as a failing test.

Asserting the returned list for `codesign: false` is a few lines and pins the comment's claim.

MAUstaoglu:
Confirmed and fixed in f7d2dbc.

Reproduced the mutation result before changing anything — against the signing suite specifically, since this machine has 58 pre-existing failures in `tvos_app_bundle_test.dart` (pbxproj script tests that need a real environment) which make a whole-suite signal unreliable:

| mutation | before | after |
|---|---|---|
| drop `CODE_SIGNING_ALLOWED=NO` | survives | **caught** |
| drop `CODE_SIGNING_REQUIRED=NO` | survives | **caught** |
| drop `CODE_SIGN_IDENTITY=` | survives | **caught** |
| drop `CODE_SIGN_ENTITLEMENTS=` | survives | **caught** |
| drop `EXPANDED_CODE_SIGN_IDENTITY=` | survives | **caught** |
| bypass the `!codesign` branch entirely | survives | **caught** |

`_resolveSigningArgs` → `resolveSigningArgs` with `@visibleForTesting`, matching the four existing such members in the file, and the new group asserts the exact returned list rather than `containsAll` — the claim in the source comment is that every one of them is required, so adding a setting to production code should make a reviewer look at the test.

Your point about *where* the failure would surface is the reason the assertion is on the resolver and not somewhere more convenient. These settings only do anything on a machine with no certificate, so the test is the only place a regression can be cheap.

## One from your table I did not fix, deliberately

**`stop suppressing the auth args when unsigned` still survives** — and I think it should, because it is redundant defence rather than an uncovered payload.

`xcodebuildArgs` already drops them:

```dart
if (!isSimulator && codesign) ...authenticationArgs,
```

and the existing `--no-codesign drops provisioning from a device build` test passes `authenticationArgs` in and asserts `-authenticationKeyID` does not reach the argv. So mutating the ternary in `build()` changes nothing observable in what xcodebuild receives.

Its one unique effect is not *resolving* the args, which avoids a spurious "App Store Connect key partly configured" warning on an unsigned build. That is real but it lives inside `build()`, behind a `processManager.run` of xcodebuild — pinning it means restructuring `build()` for a log line. Happy to do it if you disagree; it seemed a worse trade than saying so here.

## A note on the group above it

Adding the payload group made the plumbing group's value clearer, so I left a comment at the top of the new one saying which is which. Both are needed: the plumbing one would survive deleting all five settings, and the payload one would survive `xcodebuildArgs` ignoring `codesign` altogether.

Suite: 28 in `tvos_code_signing_test.dart` (was 24), analyzer clean with `--fatal-infos`.

DenisovAV (approved):
Verified against `f7d2dbc` by re-running the mutations, not from the description. 453 tests green.

| mutation | before | now |
|---|---|---|
| drop `CODE_SIGNING_ALLOWED=NO` | survived | fails |
| drop `CODE_SIGNING_REQUIRED=NO` | survived | fails |
| drop `CODE_SIGN_IDENTITY=` | survived | fails |
| drop `CODE_SIGN_ENTITLEMENTS=` | survived | fails |
| drop `EXPANDED_CODE_SIGN_IDENTITY=` | survived | fails |
| bypass the `!codesign` branch | survived | fails (−3) |
| default `codesign` to `false` | fails | fails |
| stop suppressing the auth args when unsigned | survived | **survives** |

Opening the function rather than testing through the argv builder is the right call, and the doc comment explaining *why* it is `@visibleForTesting` — that this branch is only ever exercised on a machine with no certificate — is worth more than the tests, because it is what stops someone closing it again.

The remaining survivor is diagnostics-only and I would not hold the PR for it: `resolveAuthenticationArgs` returns an empty list unless all three env vars are set, so dropping `|| !buildInfo.codesign` changes nothing functionally. What it does change is that a machine with the credentials exported would print *"Authenticating xcodebuild with App Store Connect API key …"* on an unsigned build that authenticates nothing. Same class as the simulator-guard residual on #74.

One aside on the 58 failures that made you distrust the whole-suite signal: those are the `TMPDIR` artifact, not real. With the invocation now in `CONTRIBUTING.md` —
`TMPDIR="$(cd "${TMPDIR:-/tmp}" && pwd -P)" flutter/bin/dart test test/` — the full suite is 453/453 here.

Approving.

MAUstaoglu:
Withdrawing the baseline claim from my comment above — you're right that the 58 are the `TMPDIR` artifact and not real failures.

Verified here before writing this:

```
TMPDIR="$(cd "${TMPDIR:-/tmp}" && pwd -P)" flutter/bin/dart test test/
00:06 +453: All tests passed!
```

453/453, matching your count. So "58 pre-existing failures in `tvos_app_bundle_test.dart` (pbxproj script tests that need a real environment)" is wrong on both halves: they are not pre-existing, and they do not need a real environment. `/var` is a symlink to `/private/var`, the harness's FS guard resolves symlinks on one side of its comparison and not the other, and that is the whole of it.

This is the trap `8ad2129` was written to close five days ago. CONTRIBUTING.md has carried the resolved-`TMPDIR` invocation and the explanation since then; I ran `dart test` without the prefix anyway, and then used the phantom failures as the stated reason to scope the mutation runs to the signing suite rather than the full one.

The scoping did not change what the mutation table shows — those six are the mutations the finding was about, and you re-ran them against `f7d2dbc` independently — but the claim about the suite should not stand in the thread.

PR #77, opened 2026-09-08
Point the SPM note at the build phases, not at a missing file

MAUstaoglu:
`doc/architecture.md` sends the reader to a section of `CLAUDE.md`, which is globally gitignored and so exists in nobody's clone — a dead reference in shipped documentation. 1740aab dropped two references like it and missed this one.

It now names the "Prepare Flutter framework" pre-action and the "Embed Flutter framework" build phase in the Runner project, which a reader can actually go and look at.

Doc-only; no code paths touched.


MAUstaoglu:
Folding this into #76 instead.

PR #79, opened 2026-09-10
Bump flutter-tvos to Flutter 3.47.3

MAUstaoglu:
The release train's upgrade to Flutter `3.47.3` (flutter-tvos `1.10.2`).

Everything passed — the engine artifacts published, the CLI analyzes and tests
clean against the new SDK, and the end-to-end gate is green — but `dev`
moved while the gate was running, so this could not be fast-forwarded and the
train stopped rather than deciding what the history should be.

Merging it hands over to the existing automation: Test, Sync main, Tag release,
and then `v3.47.3-tvos.1.10.2`.


DenisovAV (approved):
The bump is sound and merging is safe. Verified rather than taken from the body:

- `e8113bf45620…` is exactly the `3.47.3` tag in `flutter/flutter`. The delta from `3.47.2` includes a Dart SDK move (`Update Flutter DEPS to Dart 1d1a730e…`), which makes the engine rebuild mandatory rather than optional — AOT snapshots are keyed to the Dart hash.
- `engine-87021d7a…` exists, `isDraft: false`, all six zips plus manifests uploaded. All six manifests agree on `engine_sha` and `flutter_commit`, and the commit is confirmed *in the binaries* — each framework's `Info.plist` carries `FlutterEngine = e8113bf4…` with the right `BuildMode`.
- Platform identity is correct in all four tvOS zips: device builds report `TVOS`, the simulator build `TVOSSIMULATOR`, all arm64, all `sdk 26.5`. No variant drifted — `args.gn`, `LC_BUILD_VERSION` and `ClangVersion` agree across all six, so the six machines did not roll mid-run.
- `dart analyze lib/ test/ bin/` clean, `453` tests pass.
- `1.10.1 → 1.10.2` matches precedent: every previous "Flutter patch release + six variants rebuilt" was a patch bump (1.3.1, 1.3.2, 1.3.3, 1.4.1). The one minor bump, 1.7.0, was a Flutter *minor* move.
- `pubspec.lock`'s two moves match flutter_tools' own pins at 3.47.3 exactly. Not a loose resolve: `meta` stayed at 1.18.3 even though the override would admit 1.19.0.

What follows is about the train, not the diff.

## The stated reason for stopping is false

> `dev` moved while the gate was running, so this could not be fast-forwarded

It did not move.

```
dev tip            a4555555ed79   2026-09-09T13:57:17Z  "Merge pull request #76"
075b9d3a parent    a4555555ed79
compare dev...075b9d3a  →  ahead_by=1, behind_by=0, merge_base=a4555555ed79
```

`dev` last moved **24½ hours before the train made its commit**, and the bump is a textbook one-commit fast-forward on its current tip. The push cannot have been refused for non-fast-forwardness.

The real cause is visible in the branch settings:

```
required_approving_review_count: 1
enforce_admins: true
required_status_checks: none
```

`dev` requires a reviewed pull request, and admin enforcement is on, so **no token can push to it directly**.

This matters beyond the wording. It is the same class of defect that #7 removed from `engine-upgrade.sh` — one hardcoded diagnosis standing in for every possible push failure. There the fix was to stop discarding git's reason; here `promote.sh:132-133` correctly leaves stderr alone, so the true reason is in the build log, but the two artifacts a human actually reads in the morning — this PR body and issue #80 — still assert the wrong one. And this explanation is precisely what would talk a reviewer out of looking.

## The consequence is larger: `promote`'s success path is unreachable

`promote.sh:134-140` pushes to `dev`, then clears the attestation and deletes the branch. Given the protection above, that push can never succeed, so **the two cleanup lines have never executed**. The design premise stated in the script header and `README:187-197` — *"No pull request, deliberately… The gate is the review"* — contradicts the repository's actual configuration.

The leak is already visible, and it is two runs deep:

```
release-train:  refs/heads/train/gate-075b9d3abcd35fa0766e34a2684a71a66e965c21
flutter-tvos:   refs/heads/train/flutter-3.47.3
                refs/heads/train/no-codesign      ← from PR #76, merged 9 Sept
```

`promote.sh:136-137` states the hazard itself: *"a stale attestation is worse than none: it would let a later hand-started promote of this same commit through."* That is the state the repo is in now, and it will not self-correct — once #79 merges, a re-run exits at the idempotency check before ever reaching the clear.

So the PR fallback is not the rare branch. It is the only branch. Worth rewriting it as the normal path — clearing the attestation and deleting the branch there — or granting the train a bypass and keeping the current design.

## Two claims in the CHANGELOG that overstate

**"the tvOS engine unit tests passed."** Only `tvos_debug_sim_arm64` records `tests_ran: true`. The other five — both device AOT variants and both host variants — are `false`. Running unit tests on the simulator variant only is a reasonable choice; the sentence does not say so. And the manifest records that tests *ran*, never that they passed.

**The whole success paragraph is a template, not a measurement.** "the patch set applied and the tvOS engine unit tests passed, the CLI analyzes and tests clean, and the end-to-end matrix ran…" is a hardcoded string in `cli-bump.sh`, written into the commit before `cli-bump` triggers the gate. It is emitted on every run, including runs whose gate later fails. It happens to be true here — the gate genuinely passed, and the attestation ref is the evidence — but the text is not what establishes it.

## Smaller

`detect.sh:19-44` still documents the five hand-picked DEPS inputs and says "Nothing else in DEPS is consulted". The code at `:162-186` now fingerprints 107 deps plus 23 vars. That block is also what `--help` prints.

The gate ran `sim-debug,compile-profile,compile-release` with `--device skip`. Profile and release were compiled unsigned and never executed, so on-device JIT, runtime AOT platform identity and a signed install are outside what "green" covers here. That is by design and stated in the script — worth knowing it is what the claim excludes.

Approving the bump.

PR #81, opened 2026-09-12
Test a release-train bump on its own branch

MAUstaoglu:
The release train lands a version bump by fast-forwarding `dev` onto
`train/flutter-<version>` directly. No pull request is opened on that path, so
the `pull_request` trigger never fires for a bump commit and it reaches `dev`
carrying no checks at all.

This was invisible while the train could not push. 3.47.3's bump does have an
`analyze` and a `test` run against it — but only because the push was refused
by branch protection and the fallback opened #79 three seconds later. On the
path the train is designed to take, the bump commit would have had none.

Adding `train/**` to the push triggers puts both checks on the bump commit
before anything proposes moving `dev` onto it, and makes that commit eligible
for a required-status-check rule on `dev` — a rule that would otherwise reject
every push the train makes, since the commit would carry no checks to satisfy
it.

Only the trigger list changes; the jobs are untouched.


DenisovAV (approved):
The change is right, and the reasoning in the description is the correct one: on the path the train is designed to take, no pull request is opened, so the `pull_request` trigger never fires and a bump would reach `dev` with no checks at all.

One thing it depends on that isn't in place yet.

`dev` currently has no branch protection:

```
$ gh api repos/fluttertv/flutter-tvos/branches/dev/protection
{"message":"Branch not protected", … "status":"404"}
```

Three days ago it required one approving review with `enforce_admins: true` — which is what made the train fall back to #79. So the protection was removed to let the train push, rather than the train being given a bypass.

That fixes the train and opens `dev` more widely than it needs: anyone with write access can now push directly, with no review and no required checks, for any change — not only bumps.

This PR is the half that makes the other half possible. The description already points at it: putting checks on the `train/**` commit "makes that commit eligible for a required-status-check rule on `dev`". But that rule doesn't exist, so today the new runs are informational — a red `test` on a train branch would not stop anything from reaching `dev`.

The shape that gets both is protection back on `dev` with `analyze` and `test` as required status checks, and the review requirement bypassed for the train's identity only. The train's pushes then pass because this PR gives them the checks; everyone else's direct pushes are still refused.

Approving the trigger change — the protection is a settings change, so it can't ride in this PR.

Issue #84, opened 2026-09-14
Xcode 27: check tvOS Device Support layout and the lldb sysroot


MAUstaoglu:
Follow-up from the review of #82. Not blocking anything today: this is about Xcode 27, which is in beta.

## What changed for iOS

[flutter/flutter#189284](https://github.com/flutter/flutter/issues/189284): Xcode 27 stores iOS Device Support symbols one level deeper, at `<device>/arm64e/Symbols` instead of `<device>/Symbols`. lldb's own lookup misses them, and launch hangs for minutes. Upstream's fix, shipped in Flutter 3.47.4:

- `IOSDeviceSupport.existingDeviceSupportSymbols` prefers `<device>/<arch>/Symbols`.
- `LLDB.attachAndStart` sends `platform select remote-ios --sysroot "<symbols>"` before attaching, so lldb is told where the symbols are instead of finding them itself.

## Where tvOS stands

`TvosDeviceSupport` returns `null` from `existingDeviceSupportSymbols`, so lldb is never given a sysroot for an Apple TV and keeps relying on its own lookup, the one that broke for iOS. That was deliberate: the command is hardcoded to `remote-ios` in a private method (`_addSymbolSearchPaths` in `lldb.dart`), and selecting the iOS platform for a tvOS process is untested.

What #82 already covers: the slow-attach warning treats a `.finalized` copy with symbols one level down as complete, so it will not tell a user that Xcode "has not finished" if tvOS adopts the new layout.

What it does not cover: if lldb misses tvOS symbols the way it misses iOS ones, the attach runs into `FLUTTER_TVOS_LLDB_ATTACH_TIMEOUT_SECONDS` (180 s by default) and falls back to the Xcode debugger, which does not reliably sustain a tvOS debug session.

Nothing shows yet that tvOS gets the same layout. #189284 is iOS-only, and devicectl reports an A12 Apple TV as `arm64`, not `arm64e`.

## What would settle it

1. On the Xcode 27 beta, with an Apple TV on the matching tvOS beta, let Xcode prepare the device and record the layout under `~/Library/Developer/Xcode/tvOS DeviceSupport/<model> <version> (<build>)/`: `Symbols` directly, or `<arch>/Symbols`, and which arch.
2. Run `flutter-tvos run -d <id> --debug -v` and note whether lldb logs `libobjc.A.dylib is being read from process memory`, and how long the attach takes.

[...]

PR #86, opened 2026-09-20
Bump flutter-tvos to Flutter 3.47.5 and the A8-safe engine

### LINKED ISSUE #85: Flutter 3.47.5: the CLI needs porting
The release train re-pinned flutter-tvos to Flutter `3.47.5` (`6a19cca56475dbfba1478ee68d7bd0c2ef891da1`)
and the CLI does not compile or start against it:

```
lib/tvos_device.dart:829:41: Error: Required named parameter 'deviceVersion' must be provided.
        final LLDB lldb = _lldb ??= LLDB(
                                        ^
flutter/packages/flutter_tools/lib/src/ios/lldb.dart:25:3: Context: Found this candidate, but the arguments don't match.
```

`flutter_tools` is not a public API; a required parameter added upstream is the
usual cause. The port has to land together with the re-pin, because it will not
compile against the old Flutter.

Reproduce:

```bash
git clone --branch dev https://github.com/fluttertv/flutter-tvos.git
cd flutter-tvos
printf '%s' 6a19cca56475dbfba1478ee68d7bd0c2ef891da1 > bin/internal/flutter.version
printf '%s' engine-28b77ed94b4cb0dbb7dd64bfed02ad12d6113c0f > bin/internal/engine.version
./bin/flutter-tvos --version
```

Nothing was pushed. The train is stopped until the CLI carries to `3.47.5`.

MAUstaoglu:
Closes #85.

The train re-pinned to Flutter 3.47.5 this morning and `cli-bump` stopped at compile:

```
lib/tvos_device.dart:829:41: Error: Required named parameter 'deviceVersion' must be provided.
```

The port cannot land ahead of the re-pin (the parameter does not exist in 3.47.4), so this carries both, in the shape the train produces: 1.10.3 → 1.10.4.

## The port

3.47.5 added `required Version? deviceVersion` to the `LLDB` constructor. From 27 on it sets the JIT breakpoint without `auto-continue` and drives lldb's stops by hand ([flutter/flutter#192810](https://github.com/flutter/flutter/pull/192810)).

`TvosDevice.tvosVersion` reads the version segment of what devicectl reports (`tvOS 26.6 23L773` → `26.6`) and `lldbForDebugSession` hands it over.

What that buys on an Apple TV is not upstream's reason. Upstream's crash is a JIT breakpoint that fails to stop; on this engine it never fires at all, because `patches/01-dart-globals.patch` maps JIT pages RWX up front and stubs `NOTIFY_DEBUGGER_ABOUT_RX_PAGES` out. What `>= 27` changes here is that lldb's `detach` stop hook is dropped and crash handling moves into the tool's stdout listener, which matches line wording taken from an iOS transcript. Nothing here runs tvOS 27, so that path is unverified; #84 is where it gets a run. Reporting the real version is still right — it is what upstream asks for, and it is what will be correct when tvOS does take that path.

## The engine

`bin/internal/engine.version` moves to [`engine-e149d9fdc27e`](https://github.com/fluttertv/engine-artifacts/releases/tag/engine-e149d9fdc27ee15f0ebec71bc65ec278d5a1417c), which carries two fixes that no Flutter release would have pulled in on its own:

- **[fluttertv/engine#22](https://github.com/fluttertv/engine/pull/22)** — instanced draws no longer go through the base-vertex/base-instance Metal selectors, which need an A9. The Apple TV HD is an A8, so every `flutter_gpu` draw with an instance count aborted the app there. That is every `flutter_scene` app, and App Review tests on an Apple TV HD. Apps already shipping need a rebuild against this release.
- **[fluttertv/engine#20](https://github.com/fluttertv/engine/issues/20)** — all six zips now ship Flutter's `LICENSE` and the third-party notices aggregate. The tvOS zips shipped buildroot's C++ header comment naming Chromium; the host zips shipped nothing.

The engine published this morning (`engine-28b77ed…`) sits one commit above the #20 merge, so it carried the notices and lacked only #22. It has been deleted.

## Verified

Against Flutter 3.47.5 and the new engine, at `5afa3a2`:

- `verify_artifacts.sh`: every check, including the new A8 selector check and the notices in all six variants.
- `dart analyze --fatal-infos lib/ test/ bin/`: no issues. `dart test test/`: 473 passed.
- Mutation-checked: `deviceVersion: null` at the call site, a wrong constant, the unanchored parse and a major-only parse each fail the suite.
- **Simulator**: sim-debug 10/10 — launch, VM service, hot reload and restart each taking effect, platform identity, 10 FFI symbols.
- **Apple TV 4K (AppleTV14,1, tvOS 26.6, Xcode 26.6)**: device-debug 10/10 (hot reload 434 ms, hot restart 2.2 s, both verified in the running app), device-profile 5/5, device-release 4/4. 23 passed, 0 failed.
- The debug run attached through lldb with no fallback to the Xcode debugger, and printed no Device Support warning.

Not covered: no Apple TV HD here, so #22's fix is verified by the engine's own selector check and by App Review approving a build linked against it, not by a run on A8 hardware.

Merging releases `v3.47.5-tvos.1.10.4`.


DenisovAV (commented):
Reviewed at `284c013` against upstream 3.47.4/3.47.5 and the published artifacts.

The engine side checks out. I downloaded all six zips of `engine-e149d9fd…` and ran the engine's own `verify_artifacts.sh` at `e149d9fd`: 28 checks, all pass, including 3b and the notices in every variant. The same script against the 3.47.3 release fails exactly where this one fixes things (`MISSING the A8 instanced-draw patch`, `LICENSE is a source-file header`, `host_release ships no LICENSE`), so the checks discriminate. Independently: the plain instanced selectors are present and `…vertexCount:instanceCount:baseInstance:` is gone in all four tvOS variants — in `libFlutter.dylib` and the xcframework slices too, not just the binary check 3b reads; `LICENSE` is byte-identical to upstream `engine/src/flutter/LICENSE` at `6a19cca5`; platforms are 3/3/3/8. All six manifests record `flutter_commit` `6a19cca5…` and `engine_sha` `e149d9fd…`, and the shipped `args.gn` carries `dart_version = b530c21f…`, which is what 3.47.5's DEPS names. `LLDB` is the only construction site here that gained a required parameter, and nothing else in the 3.47.4 → 3.47.5 diff needs a reaction. The tag resolves to `v3.47.5-tvos.1.10.4` and the CHANGELOG section extracts cleanly.

The rest is about the port.

## The reason given for passing the version does not hold for this engine

The description, `CHANGELOG.md:35-39` and `tvos_device.dart:370-373` all say `null` would take an Apple TV on tvOS 27 down the path that crashes the app. Upstream's crash is a JIT breakpoint that fails to stop under `--auto-continue`, leaving the RX pages untouched. On a physical Apple TV this engine never gets there: `patches/01-dart-globals.patch` sets `FLAG_write_protect_code = false` in `VirtualMemory::Init`, so JIT pages are mapped RWX up front, and it defines `NOTIFY_DEBUGGER_ABOUT_RX_PAGES` as an empty stub whose own comment says tvOS "never calls this hook" and that the symbol exists only so the breakpoint resolves and the attach completes. A breakpoint that never fires cannot crash the app for failing to stop.

What passing ≥ 27 does do on tvOS is drop the stop hook: upstream installs `target stop-hook add -o "thread backtrace all" -o "detach"` only when `manualContinue` is false (`lldb.dart:189-193`). With it on, crash handling moves into the stdout listener (`:245-288`), which backtraces and detaches only after seeing `Process N stopped` … `Target 0: (X) stopped.` without `stop reason = breakpoint` in between — shapes upstream pins against an iOS transcript. If tvOS lldb words either line differently over the CoreDevice tunnel, the debugger stays attached and the app hangs with no backtrace, and on tvOS the detach is what releases it.

I'd still pass the version — it is what upstream asks for, and it is right the moment tvOS does use that hook — but the claim should say what it actually buys, and the stop-hook trade is worth a line. This is unverifiable without tvOS 27 hardware, so #84 is the natural place for the run. The CHANGELOG text ships as the release body on merge, which is why it's worth fixing first.

## The line this PR exists for is not pinned by a test

Six mutations, full suite each, baseline 470 with only the known `tvos_create_test` environment failure:

| mutation | result |
|---|---|
| `deviceVersion: null` (`tvos_device.dart:849`) | survives |
| `deviceVersion: Version(26, 0, 0)` | survives |
| parse from `osVersion` only | survives |
| parse from `deviceSupportVersion` only | survives |
| reject versions with no build number | survives |
| regex `\d+` (major only) | killed |

`dart analyze --fatal-infos` is clean on the `null` mutant too, so both CI gates pass with the one line reverted. The tests assert `TvosDevice.tvosVersion` and nothing downstream of it. A test that asserts what `startApp` hands `LLDB` — `_lldb ??=` at `:845` takes an injected instance — plus one case per source would close it.

## A bare build number parses as a version

`tvos_device.dart:376` takes the first digit run anywhere in the string, and `tvos_emulator.dart:201-205` joins only the halves it has. With `osVersionNumber` missing and `osBuildUpdate` present, `osVersion` is the bare build (`23L773`) and `tvosVersion` becomes `Version(23, 0, 0)` — a silent sub-27 answer on what may be a tvOS 27 device, which is the case this change exists to get right. A non-numeric version does the same (`tvOS Unknown 24A123` → 24). Anchoring on the version segment (`^(?:tvOS )?(\d+(?:\.\d+)*)`, or keeping `osVersionNumber` as its own field) fixes both.

Related: `?? deviceSupportVersion` is unreachable. `tvos_emulator.dart:215-217` sets `deviceSupportVersion` only when both the version and the build are present, and in that case `osVersion` is non-null too — so the doc comment at `:365-367` ("on a device discovered without a build number") describes a state that cannot occur.

## Minor

`pubspec.lock` still pins `dds` 5.3.0 and `devtools_shared` 12.1.0; 3.47.5 moved them to 5.4.0 and 13.1.0. CI runs a plain `pub get`, so nothing fails, but `bin/internal/shared.sh:198` tries `pub get --offline` first, which will now miss on a cache built from the lock.

The description says the engine published this morning, `engine-28b77ed…`, "predates both". It sits one commit above `c317120`, the #21 merge, so it carries the notices and lacks only #22 — which is how fluttertv/engine#24's own description puts it. That release also no longer exists (`releases/tags/engine-28b77ed…` is a 404, drafts included), so "will be removed once this ships" is already done.

MAUstaoglu:
All four check out against the code; fixed in 5afa3a2.

**The reason for passing the version.** You're right, and I'd had it backwards. `patches/01-dart-globals.patch` sets `FLAG_write_protect_code = false` in `VirtualMemory::Init` and defines `NOTIFY_DEBUGGER_ABOUT_RX_PAGES` as an empty stub whose comment says tvOS "never calls this hook", so upstream's breakpoint cannot fire, let alone crash the app for failing to stop. What `>= 27` actually changes here is `lldb.dart:189-193`: the `detach` stop hook is dropped and crash handling moves into the stdout listener at `:245-288`, which keys off `Process N stopped` and `Target 0: (X) stopped.` — iOS wording, unverified over the CoreDevice tunnel, and on tvOS the detach is what releases the app. The CHANGELOG, the `tvosVersion` doc and the description now say that and point at #84. The version still gets passed, for the reason you give.

**The untested line.** Confirmed: `deviceVersion: null` passed the suite and `analyze --fatal-infos`. The construction moved out of `startApp` into `lldbForDebugSession`, with the constructor behind a replaceable `lldbFactory`, so a test can observe what is handed over. Re-running your table on the new tests:

| mutation | result |
|---|---|
| `deviceVersion: null` | caught |
| `deviceVersion: Version(26, 0, 0)` | caught |
| unanchored parse | caught |
| major-only parse (`\d+`) | caught |

The test also asserts the instance is created once and reused.

**The bare build number.** Reproduced: `23L773` gave `Version(23, 0, 0)`. The parse is anchored now, with a guard that rejects a version run butted against letters (`^(?:tvOS )?(\d+(?:\.\d+)*)(?![\dA-Za-z])`) — an anchor alone was not enough, since the build starts with digits. Covered for the bare build, `tvOS 18.4` from a simulator runtime, and a runtime with no number (`tvOS`) at all. The unreachable `?? deviceSupportVersion` and the doc sentence describing it are gone.

**`pubspec.lock`.** Updated to 3.47.5's `dds` 5.4.0 and `devtools_shared` 13.1.0, so `shared.sh`'s `pub get --offline` doesn't miss.

**The description.** Corrected: `engine-28b77ed…` carried the notices and lacked only #22, and it is already deleted, not pending.

473 tests, analyzer clean. The device matrix was run at `284c013`; 5afa3a2 changes the parse, the factory seam and text, with no change to what the attach does at 26.6 — happy to re-run it on the Apple TV if you'd rather have the numbers from this commit.

MAUstaoglu:
Re-ran the full matrix at `5afa3a2`, this time with the unit phase in: **33 passed, 1 failed, 1 skipped.**

- Simulator debug 10/10; Apple TV 4K (tvOS 26.6, Xcode 26.6) device-debug 10/10 — hot reload and hot restart both taking effect — device-profile 5/5, device-release 4/4.
- `analyze` clean, `dart test test/` 473 passed.
- Debug attached through lldb with no fallback to the Xcode debugger, and no Device Support or slow-attach message.

The one failure is the harness, not the build: `verify_artifacts` runs before `precache`, and this run started from a checkout whose `engine_artifacts/` was empty, so it had nothing to inspect ("artifacts dir not found"). The scenarios that follow downloaded the engine and used it. Running the same script against the populated directory afterwards passes all 28 checks, including the A8 selector check and the notices in every variant.

DenisovAV (approved):
Re-checked at `5afa3a2`. Baseline here is 473 with only the known `tvos_create_test` environment failure, analyzer clean, and `pubspec.lock` now on `dds` 5.4.0 / `devtools_shared` 13.1.0. Re-ran the table against the new tests:

| mutation | result |
|---|---|
| `null` for `tvosVersion` at the factory call (`tvos_device.dart:400`) | caught |
| a constant `Version(26, 0, 0)` there | caught |
| unanchored parse | caught |
| parse without the trailing guard (`23L773` → 23) | caught |
| major-only `\d+` | caught |

The rationale now reads as what the change actually buys, in the description, `CHANGELOG.md:16-26` and the `tvosVersion` doc, and the `engine-28b77ed…` sentence is right.

Two things the new seam leaves open, both small:

**The mapping into `LLDB` is still unpinned.** `_newLldb` (`tvos_device.dart:406-417`) is the one place the parameter reaches the constructor, and the tests replace `lldbFactory` before it runs. Changing `deviceVersion: deviceVersion` to `deviceVersion: null` there leaves the suite at baseline. The seam pins `lldbForDebugSession` → factory; factory → `LLDB` is what stays unobserved, which is the same boundary as before, one call deeper. Closing it properly means what upstream does in `lldb_test.dart` — drive a real `LLDB` through `attachAndStart` against a fake process manager and assert the breakpoint command shape, `--auto-continue` present below 27 and absent from 27 — which is more machinery than this PR needs. Worth knowing it is not covered.

**"Created once, then reused" is not what that assertion tests.** The fake factory returns the same instance on every call (`tvos_device_support_test.dart:253-258`), so both `expect`s pass whether or not `_lldb ??=` caches: dropping the cache leaves the suite at baseline. Counting factory calls would test it.

Approving — neither blocks the release, and the matrix at `5afa3a2` covers what a user hits on 26.6.

One note for the gate rather than this PR: the run's single failure came from `verify_artifacts` executing before `precache` against an empty `engine_artifacts/`, which reports "artifacts dir not found" as a failure. A scenario that depends on a populated cache either wants the download ahead of it or a skip that says so; as it stands a fresh checkout fails a check that nothing is wrong with.

